PR #5, opened 2018-03-18
Fixes Installation Error Due To Wrong Directory Structure

### LINKED ISSUE #4: Installation using pip install  gives FileNotFoundError
The following error is encountered while installing with `pip install .`. 
image

This is because setup.py expects a directory structure as `lib/hydra` whereas the structure in the current repository is just `hydra`. 
https://github.com/HTTP-APIs/python-hydra-agent/blob/925f6277649eb16087c4c95ebea3aa2c3011f9d6/setup.py#L11 

Can be solved by changing directory structure and modifying the `gitignore` to include `lib/hydra`

suryatejreddy:
Fixes #4 

I changed the directory structure from `hydra` to `lib/hydra` and removed `lib/` from `.gitignore`. 

Now the module can be installed using `pip install .` .


Mec-iS:
`lib` was the old directory but I removed that layout.

Try now on `develop`:
https://github.com/HTTP-APIs/python-hydra-agent/blob/develop/setup.py#L11

Mec-iS:
https://github.com/HTTP-APIs/python-hydra-agent/commit/57da1389cebac5c6eea60a2c69b0288e6fe697b7

PR #7, opened 2018-03-25
Added missing dependency of hydra lib used for testcases.

anubhav253:
Hydra module was missing here in the requirements list which was being used by the test cases


ashwani99:
@anubhav253 `hydra` is the name of the package itself, its not a dependency.
Also, it seems that you are trying to merge a copy of master branch into develop branch. First, you need to checkout to develop branch, create another branch for making changes, make changes and send PR for develop branch.

anubhav253:
Hi @ashwani99 if someone would try to run hydra on his local repo then we will need this file as its a dependency for the testcases. you might not be getting this error because you already had hydra lib installed in your pc. I got module not found error on a vm.
Check this usage here. 
https://github.com/HTTP-APIs/python-hydra-agent/blob/925f6277649eb16087c4c95ebea3aa2c3011f9d6/examples/test0.py#L7

Regarding merging master branch to develop we can always do that since master here for the repos is the parent branch for the develop. A new commit that you see here was because of some other changes that i did which was already to be updated with the upstream develop but was not available in my remote develop. Thats why its was merge commit without any merge changes

ashwani99:
Hey @anubhav253 You will need to install the hydra package first, go to the base directory and run `python setup.py install` to install the hydra package. It's not a dependency, it's the package itself.

anubhav253:
@ashwani99 - Please find below the steps to reproduce the issue on your local machine

- pip uninstall hydra
- now run python server.py
- And try running python examples/test0.py

ashwani99:
@anubhav253 Where did you find `server.py`? Also, why are you uninstalling `hydra` before running `test0.py`?

anubhav253:
@ashwani99 When you are working on a docker or virtual machine then it will not be have the hydra library so we have to install that. Above that was a typo it is setup.py .

xadahiya:
Closing #7

PR #13, opened 2018-05-21
rough implementation of graph

sandeepsajan0:
redis_graph.py is the rough implementation of the graph using Redis. It is not a generalized implementation, It will be generalized within 1-2 days. redis_graph.py stores the graph of apidoc in Redis.   For now, It is implemented by considering only [this doc](http://www.markus-lanthaler.com/hydra/event-api/ ), there is a need to add and update some functions for the general structure. File redis_usage have instruction to creating the environment for redis_graph.py.


sandeepsajan0:
Now, the code is updated and it follows pep8 guidelines. I will write more comments later because of work on this file is continued. Please, can you provide me a file or other thing which should contain the information about the whole APIdoc which will be very helpful in generalizing the graph by adding or updating the functions of redis_graph.py?

sandeepsajan0:
Now, the proper node key has been added to the nodes and now it is easy to see and understand the graph with the help of edges and nodes. Edges are able to create a whole structure of the graph.
Now I will add some more functions for some of the other types of APIdoc and then we will bw able to get the whole graph.

chrizandr:
From what I can see(maybe I understood it wrong), you are trying to parse the whole API Documentation of a server and storing it in Redis correct? If that is the case then that would not be what we would like. The point of the storage is that we only store the API structure, you need not store all the information that is there in the API Doc in Redis, just the classes, properties, objects and their interrelations(in the form of a graph structure). I will give an example:

Suppose you have the following API with these classes:
```
class User:
    def init(self):
         self.name = Name()
         self.email = Email()
         self.friend = User()

class Name():
    def init(self):
         self.value = String()

class Email():
    def init(self):
         self.value = String()
```
This is a simple graph. How it is implemented in Hydrus is:
`User` class is stored in the `RDFClass` Table.
`Name` property is stored in the `Property` Table.
`Email` property is stored in the `Property` Table.

Suppose you have an instance of the class:
```
User : {
   "@id": 7,
   "name": "Chris",
   "email": "something@mail.com",
   "friend": "user/10"
}
```
We have an instance table that shows instances of the classes:

ID  Type
7    "User"
11  "Name"
15  "Email"
27  "User"

The graph connecting these instances is stored in the Graph table as triples:
```
ID    Subject    Predicate    Object
xx      7             "name"        11
xx      7             "email"        15
xx      7             "friend"        27
```

This is how objects of these classes are stored. During retrieval, we retrieve all properties of a given instance and display them collectively, but the problem here would be that we need to do multiple queries to get one instance of a class. When we have queries in bulk it becomes slow and inefficient. That is why we want to minimize the querying on the server side, so we cache the API structure(classes, properties) and the data in the API (instance, graph) on the client side using Redis. 

I hope this clears it up a bit. I haven't done a detailed review of the code yet, it would be great if you could provide a README explaining what is happening. Also code is still not PEP8 compliant. One suggestion is to use flake8 to check if your code follows PEP8 or not.

You can ignore this if your code is actually doing all that, like I said I haven't been able to do a detailed review, just putting it here for reference for now.

xadahiya:
I agree with Chris, you don't need to store the entire APIDoc in Redis. We just need to extract the API structure from APIDoc then in order to retrieve and cache the data you'll need to recursively call those data endpoint and their respective context endpoints ( In order to fetch the details of data stored at every endpoint you visit. It may be a collection endpoint or a simple one). Keep in mind why we're doing this i.e to create an efficient querying process without increasing much load on the HYDRUS server.
I think first you should start a HYDRUS instance and then use hydra-py to fetch data and the API structure from server in order to understand how things work in general.

sandeepsajan0:
I am trying to understand where I lack your point. For this, I have an implementation idea here https://docs.google.com/document/d/1X3XuTxqw4eLsif9XObPt9MxMyxZrpP1r-xFSt1r82Kg/edit
Is it close, to that you want to say? And please correct and suggest me there with the example.

sandeepsajan0:
For now, I have created a new file(redisgraph_demo1) in which I am trying to store only API structure in the graph and I also add a readme for this implementation. Please tell me this implementation idea(readme) is fine or not? If not please suggest the changes.

vaibhavchellani:
@sandeepsajan0 you can try [this](http://pep8online.com/checkresult) to check

hydra/graph_readme.md
```diff
@@ -0,0 +1,59 @@
...
+ if we get the first endpoint apidocumentation the we will go on apidocumentation node show its properties like this:
+        apidocumentaion {
+                            title : "hydra:title"
+                            description : "hydra:description"
+                            entrypoint : "hydra:entrypoint"
+                            supportedclasses : "hydra:supportedclasses"
+                        }
+  and if now client have query for supportedclasses of the same apidocumentaion_id(above) then we will move forward in graph go to the node supportedclasses and will show all titles of it like:
+        supportedclasses {
+                             title1: "class_title1"
```
xadahiya: You can't just use supported_classes[title] to extract the API structure or query endpoints. Take a look at this event API documentation example http://www.markus-lanthaler.com/hydra/event-api/vocab . You'll need to use vocab:entrypoint and probably some recursive technique to extract the endpoint information like URL. First see the event API example and try to figure out how can you extract the endpoint information for APIDOC then later you can generalize this.

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,274 @@
...
+    create_graph(node, node1, "has_supportedoperation")
+
+
+def create_graph(source_node, dest_node, predicate):
+    edge = Edge(source_node, predicate, dest_node)
+    redis_graph.add_edge(edge)
+
+
+apidocumentation()
+
```
xadahiya: Hey sandeep, instead of just storing the whole documentation in Redis graph. Try to visualize things first. For example first you should think about a rough graph structure that can be easily used to store the complete api structure along with what kind of data needs to be stored at each endpoint. Forget the APIDoc, try to visualize this first for a simple API you know. Then try to find a way to extract all that required information from APIDOC and after that you can retrieve data from the server and add it to the graph for querying. Makes sense?

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,274 @@
...
+    node = Node(
+        label="node",
+        alias="status_code",
+        properties={
+            "prop": str(status_code)})
+    redis_graph.add_node(node)
+    create_graph(status_node, node, "has_status")
+
+
+def Property(property_node):
```
xadahiya: Please add proper description annotations to each function using `""" """`. Also, make sure your code is PEP8 complaint.

xadahiya:
selection_019
 Getting this error when trying to run the code.

xadahiya:
selection_020
What are these members? Also, I'm getting this error  `redis.exceptions.ResponseError: unknown command 'GRAPH.QUERY'` when I use python2 instead of python3
`

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,258 @@
...
+        endpoint_member["end_date"] = str(new_dict["end_date"])
+        endpoint_member_alias = str(get_member_alias(member["@id"]))
+        print endpoint_member_alias
+        node1 = Node(
+            label="endpoint_member",
+            alias=str(endpoint_member_alias),
+            properties=endpoint_member)
+        redis_graph.add_node(node1)
+        edge = Edge(class_node, "class_get_member", node1)
+        redis_graph.add_edge(edge)
```
xadahiya: Please try to make your code modular and reusable by breaking it into smaller functions. The practice you're following is a really bad one.

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+DEFAULT_HTTP_CLIENT = Http(_MemCache())
+http = DEFAULT_HTTP_CLIENT
+APIDOC_RE = regex(
+    r'^<([^>]*)>; rel="http://www.w3.org/ns/hydra/core#apiDocumentation"$')
+
+
+def final_file(url):
+    response = urllib.urlopen(url)
```
xadahiya: Please provide proper Docstrings for each function. Read this for example http://epydoc.sourceforge.net/docstrings.html

hydra/redisgraph_demo1.py
xadahiya: Here are some practices you should try to follow https://stevenloria.com/python-best-practice/

sandeepsajan0: The above errors are occurring due to the Redis set up, I 'll provide a setup file for running the same file with Redis. And yeah I am working on it, I'll break it in reusable code functions. Actually, I  want to show that this graph can work finely? can I move forward with this graph after some required change in it?

xadahiya: I need to run your code in order to understand what it does

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,258 @@
...
+#    base_url = "/{0}/{1}/".format(match_obj.group(1), match_obj.group(2))
+    entrypoint_member = "member" + match_obj.group(4)
+    return entrypoint_member
+
+
+def remove_vocab(string):
+    return string.replace("vocab:", "")
+
+
+""" Finding the entrypoint class for extracting the endpoints"""
```
chrizandr: You did not have to do all the parsing for the API Documentation. Like I mentioned to you on the gitter chat, you can directly use the `hydrus.hydraspec.doc_maker.create_doc()` to parse the API Documentation. It will return an object of type `hydrus.hydraspec.doc_writer.HydraDoc`, you can directly access the endpoints, classes and their properties here as python objects, without having to search the dictionary for all that information. I remember asking you to read and understand the code so that you can reuse it.

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,258 @@
...
+    endpoint_list.append(endpoint["id"])
+endpoints["@id"] = str(url)
+endpoints["@type"] = "EntryPoint"
+url_node = Node(label="id", alias="url", properties=endpoints)
+redis_graph.add_node(url_node)
+
+""" Creating endpoints as the nodes and set all the properties or
+    data with the help of endpoint """
+
+for obj in entrypoint_obj["supportedProperty"]:
```
chrizandr: It is not very clear what exactly is the structure in which you are storing the API information and also not clear as to what information you are storing. As much as I can make out, you are now storing events as simple nodes and connecting all objects with a class node and the class to the collection node. Am I correct? You need to define the graph structure that you are using, maybe use a simple diagram or doc to explain what exactly is happening in the code.

hydra/redisgraph_demo1.py
```diff
@@ -0,0 +1,258 @@
...
+                    delete_classes = 1
+                if obj["method"] == "GET":
+                    get_classes = 1
+
+    """ members are leave nodes which is connect with the classes
+        every elementary operation should be perform on these.
+        like put,post,delete these operation can be done by adding and
+        deleting the nodes and update all the above nodes properties"""
+
+    for member in new_file["members"]:
```
chrizandr: Again, not able to understand at all what is happening in the code, the output simply says that some members are added to the Redis graph, please define what the graph structure is and explain how the information is stored in Redis. Also, please write give some proper output or reference so that we can view the graph as well.

hydra/redisgraph_demo1.py
sandeepsajan0: yeah, exactly I have sketched it [there](https://docs.google.com/document/d/1iQ-0UwRyrePVrmW9z7AuUyK6TiilMIimlhn9msRH3eI/edit?usp=sharing)

hydra/redisgraph_demo1.py
sandeepsajan0: yeah I'll update it.

hydra/redisgraph_demo1.py
sandeepsajan0: yeah, here I am tried to add the events(members) as the elementary nodes and store it into Redis. We can update the graph if any other event is posted or update only by adding or deleting the node and update the property in above nodes.

hydra_redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,348 @@
...
+                properties = []
+                if obj1["supportedOperation"]:
+                    objects_properties["methods"] = str(
+                        get_method(obj1["supportedOperation"]))
+                if obj1["supportedProperty"]:
+                    for obj2 in obj1["supportedProperty"]:
+                        properties.append(obj2["title"])
+                    objects_properties["property"] = str(properties)
+                    obj_properties_classlist = get_property(
+                        obj1["supportedProperty"])
```
xadahiya: You need to get all these properties from the api_doc you created in line 339.
Let me show you an example.
**You can simply use the APIDoc object to get all the required properties. No need to manually parse the object. See screenshots below**
selection_021

selection_022

selection_023

Please analyze the following files and update your code accordingly.
https://github.com/HTTP-APIs/hydrus/blob/master/hydrus/hydraspec/doc_writer.py
https://github.com/HTTP-APIs/hydrus/blob/master/hydrus/hydraspec/doc_maker.py

hydra_redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,348 @@
...
+        print(new_url)
+        response = ur.urlopen(new_url)
+        new_file = json.loads(response.read().decode('utf-8'))
+        print(new_url, new_file)
+        for obj1 in apidoc["supportedClass"]:
+            if obj == obj1["title"]:
+                obj_properties["methods"] = str(
+                    get_method(obj1["supportedOperation"]))
+                obj_properties_classlist = get_property(
+                    obj1["supportedProperty"])
```
xadahiya: Fix this as well

hydra_redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,348 @@
...
+
+def get_endpoints(entrypoint_obj):
+    print("creating entrypoint node")
+    url_node = Node(label="id", alias="Entrypoint", properties=entrypoint_obj)
+    redis_graph.add_node(url_node)
+    print("commiting")
+    redis_graph.commit()
+    print("entrypoint node ", url_node)
+    get_apistructure(entrypoint_obj, url_node)
+
```
xadahiya: You need to update all your code and use `<hydrus.hydraspec.doc_writer.HydraDoc object>` instead of manually parsing the APIDoc.

xadahiya: From the graph structure doc you provided, your graph structure looks a bit incorrect. I've suggested some changes. Please take a look https://docs.google.com/document/d/1iQ-0UwRyrePVrmW9z7AuUyK6TiilMIimlhn9msRH3eI/edit

Also, you can start with this basic structure. It'll help you with querying part as well.
img_20180602_135101

You need to generate a dynamic graph with a structure similar to this. We can deal with data redundancy later.

hydra_redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,348 @@
...
+
+url = "http://35.224.198.158:8080/api"
+match_obj = re.match(r'(.*)://(.*)/(.*)/?', url, re.M | re.I)
+base_url = "{0}://{1}".format(match_obj.group(1), match_obj.group(2))
+entrypoint = "/" + match_obj.group(3) + "/"
+print(base_url, entrypoint)
+
+
```
xadahiya: Move this under `if __name__ == "__main__":`

You can read more here http://www.bogotobogo.com/python/python_if__name__equals__main__.php

hydra_redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,348 @@
...
+def get_title(new_obj):
+    new_obj = new_obj[0]
+#    print (new_obj)
+    obj_title = new_obj["description"]
+    print(obj_title)
+    return str(obj_title[4:])
+
+
+def get_endpoint_id(string):
+    return (string.replace("vocab:EntryPoint", ""))
```
xadahiya: **As I've pointed out multiple times please use Doc Strings for every function you write.**

hydra_redis/redis_usage
```diff
@@ -0,0 +1,19 @@
...
+    
+After install redis-graph, redis and its python client 
+you should do :
+    $cd redis
+    $src/redis-server --loadmodule /path_to_redis_graph/src/redisgraph.so
+    like  this:
+    src/redis-server --loadmodule /home/sandeep/demo/redis-module-graph/src/redisgraph.so
+ it will run the redis server then you can run your redis file in different terminal but in same env.   
+
+    **docker image will provided early for now you can use the same configuration.
```
xadahiya: You should create a bash file or something to setup Redis.

hydra_redis/redisgraph_demo1.py
xadahiya: Moved the graph structure discussion to #18

hydra_redis/redis_usage
sandeepsajan0: I am trying to make a docker image for that. I'll provide it early.

hydra-redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,294 @@
...
+if __name__ == "__main__":
+    url = "http://35.224.198.158:8080/api"
+    match_obj = re.match(r'(.*)://(.*)/(.*)/?', url, re.M | re.I)
+    base_url = "{0}://{1}".format(match_obj.group(1), match_obj.group(2))
+    entrypoint = "/" + match_obj.group(3) + "/"
+    print("baseurl, entrypoint ", base_url, entrypoint)
+
+    apidoc = final_file(url + "/vocab")
+    api_doc = doc_maker.create_doc(apidoc)
+    get_endpoints(api_doc.entrypoint.get())
```
chrizandr: You are still not using the abstraction given by the `doc_maker`, when you use the `get()` method, it returns the json object, which is the same as what was being used previously.

hydra-redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,294 @@
...
+    print("commiting")
+    redis_graph.commit()
```
chrizandr: Again, we have no way of verifying what the graph is and how it is being stored, this was one of the tasks you were supposed to complete this week.

hydra-redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,294 @@
...
+
+def final_file(url):
+    """ open the given url and read and load the Json data.
+    """
+    response = ur.urlopen(url)
+    return json.loads(response.read().decode('utf-8'))
+
+
+def objects_property(objects_node, new_list, url_node):
+    """makes nodes for those properties which have properties in itself.
```
chrizandr: Docstrings usually are in one line and start with a capital letter. The qoutes should also be in the same line for a single line docstring.
`"""This is an example of a doctstring."""`

hydra-redis/redisgraph_demo1.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    print("for the properties which have class and should have node")
+    print(new_list)
+    for obj in new_list:
+        obj_method = []
+        obj_prop = []
+        properties = []
+        obj = obj.replace("vocab:", "")
+        objects_properties = {}
+        for obj1 in api_doc.parsed_classes[obj]["class"].supportedOperation:
```
chrizandr: You are using a global variable without declaring it as global, you should pass the variable as a parameter instead.

hydra-redis/redisgraph_demo1.py
xadahiya: It would be better if you follow this format https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html

hydra-redis/redisgraph_demo1.py
sandeepsajan0: In line 293, there are two commenting lines actually if you run the file with uncommenting these lines it will show the all the edges between all the nodes in the graph. So, it can be a look at the graph how it's look like and for getting the data in the graph I'll update it early.

hydra-redis/redisgraph_demo1.py
sandeepsajan0: okay, I'll update it.

hydra-redis/redisgraph_demo1.py
chrizandr: There are a lot of objects being added, possibly reduce them? We just need to make sure it works for a few different types, no need to add 30-40 objects. I suggest you try using **graphviz** for the visualisation, it is extremely simple, just a few lines of code need to be added and you can create a `.png` file showing the graph.  [example](https://github.com/pygraphviz/pygraphviz/blob/master/examples/simple.py)

hydra-redis/redisgraph_demo1.py
sandeepsajan0: Sorry, I cannot get your point here. I think, we need only the endpoints ids with its name here and it(object) provides only the same thing.

hydra-redis/redisgraph_demo1.py
sandeepsajan0: Here are some graph structures which is generated from graphviz with the code. https://docs.google.com/document/d/1h1EER1I2QXE7P2WorX7DTuUuLNIXIEqcX1ptoKXp5io/edit?usp=sharing
I have also added some lines in the code for getting the whole graph structure using graphviz.

Shekharrajak:
Spend some time to learn [SOLID](https://en.wikipedia.org/wiki/SOLID) design principles (very important) , [software design patterns](https://en.wikipedia.org/wiki/Software_design_pattern) , better exception handling, extendable & modular code and [some other concepts](http://www.informit.com/articles/article.aspx?p=2223710) to code like a pro.

sandeepsajan0:
I have updated the code regarding breaks in modules and add some more comment line and using graphviz for visualization of graph. Now I am moving forward to write test regarding this.

sandeepsajan0:
I added two tests for testing that how collection endpoints and classes endpoints are stored in Redis. And for testing the graph I have added few lines in the main file(hydra_graph.py) but still commented because if you want to test graph uncomment them. If there is a need for more test so, please suggest.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+                    obj_method.append(obj1.method)
+                print("classes ", obj_method)
+                obj_properties["methods"] = str(obj_method)
+                # store the operations for the endpoint
+                for obj1 in api_doc.parsed_classes[obj]["class"
+                                                        ].supportedProperty:
+                    properties.append(obj1.title)
+                    if obj1.prop.replace(
+                            "vocab:", "") in api_doc.parsed_classes:
+                        obj_prop.append(obj1.prop.replace("vocab:", ""))
```
xadahiya: WHY???

First of all, most of the properties won't have `vocab:` in them. They can be schema.org references lilke `'http://schema.org/geo'`. Secondly, this if will always return True.

You should use this to make the graph more connected. For example:
Suppose we have `'vocab:State'` as a prop then you know it'll an instance of `state` class and if State also happens to be an endpoint class then you can just store the URI reference to that object instead of storing the whole object. If State is not an endpoint then you need to parse `state` supportedProperties first in order to store the object.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+                obj_properties["property"] = str(properties)
+                obj_properties_classlist = obj_prop
```
xadahiya: Should be `supported_properties_list`

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+                obj_properties["class_property"] = str(
+                    obj_properties_classlist)
+                # store all those properties which is itself an object
+                for obj1 in api_doc.parsed_classes[obj]["class"
```
xadahiya: You're running the same loop twice. One at line `81` and then here line `92`.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+                                                        ].supportedProperty:
+                    if isinstance(new_file[obj1.title], str):
+                        member[obj1.title] = str(
+                            new_file[obj1.title].replace(" ", ""))
+
+            member["@id"] = str(new_file["@id"])
```
xadahiya: Create a separate function to fetch data from the server.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+            member["@type"] = str(new_file["@type"])
+            obj_properties["property"] = str(member)
+            class_object_node = Node(
+                label="classes",
+                alias=str(obj),
+                properties=obj_properties)
+            self.redis_graph.add_node(class_object_node)
+            edge = Edge(url_node, "has" + obj, class_object_node)
+            self.redis_graph.add_edge(edge)
+            # set edge between the entrypoint and the class endpoint/object
```
xadahiya: **Important** - Create separate functions for extracting properties, for fetching members from the server and storing data in Redis to make the code modular and more readable.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+from redisgraph import Node, Edge
+
+
+class Classes:
+    """Contains all the classes endpoint and the objects"""
+
+    def __init__(self, redis_graph):
+        self.redis_graph = redis_graph
+
+    def objects_property(self, objects_node, new_list, url_node, api_doc):
```
xadahiya: Explain what this function is doing. Also use a more precise Docstring. `Makes nodes for those properties which have properties in itself.` this is not very clear.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+                obj_method.append(obj1.method)
+            print("classes", obj_method)
+            objects_properties["methods"] = str(obj_method)
+            # all the operations for the object is stored in method
+            for obj1 in api_doc.parsed_classes[obj]["class"].supportedProperty:
+                properties.append(obj1.title)
+                if obj1.prop.replace("vocab:", "") in api_doc.parsed_classes:
+                    obj_prop.append(obj1.prop.replace("vocab:", ""))
+            objects_properties["property"] = str(properties)
+            # all property which itself is an object
```
xadahiya: **Important** - Again most of the code here is similar to what you wrote in `endpointclasses()`. Please reuse your code instead of rewriting it.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+            obj_alias = str(objects_node.alias + str(obj)).lower()
+            # key for the node of the object
+            print(obj_alias)
+            obj_node = Node(
+                label="object",
+                alias=obj_alias,
+                properties=objects_properties)
+            self.redis_graph.add_node(obj_node)
+            edge = Edge(objects_node, "has" + str(obj), obj_node)
+            self.redis_graph.add_edge(edge)
```
xadahiya: Write separate functions for Redis. Like there can be functions to add edge and add node etc.

hydra-redis/collections_endpoint.py
```diff
@@ -0,0 +1,161 @@
...
+#            print("commit endpoint collection")
+#            self.redis_graph.commit()
+            # create the graph in redis
+            self.collectionobjects(
+                obj_collection_node,
+                new_file["members"],
+                new_url,
+                url_node,
+                api_doc,
+                base_url
```
xadahiya: **I'm not reviewing this file right now but most of the pointers for classes_endpoint.py go for this as well.**

hydra-redis/redis_usage
```diff
@@ -0,0 +1,27 @@
+Requirements for redis_graph.py file according to my local env.
+    install redis-graph ref: http://redisgraph.io/#linux-ubuntu-1604
+    install and load module redis ref: http://redisgraph.io/#loading-the-module-to-redis
+    now for python client usage:
+    pip install redis
+    pip install redisgraph       // you can change the client.py with your requirements
+    and then first run the redis server and then run redis_graph.py file
+
```
xadahiya: Remove this file not required anymore.

hydra-redis/test/test0.py
```diff
@@ -0,0 +1,29 @@
...
+            string = obj1.decode('utf-8')
+            ob = map(str.strip, string.split(','))
+            obs= list(ob)
+            flag+=1
+        else:
+            string1 = obj1.decode('utf-8').replace("'", '"')
+            newstr = '[' + string1 + ']'
+            data = json.loads(newstr)
+            print("\nid= ",data[2]["@id"])
+            for i in range(len(obs)):
```
xadahiya: In tests you're supposed to use `unittest` or some other Python module.
**Important** - In tests we use `assert` not print. Did you read the resources @chrizandr  shared on the Gitter channel?
http://pythontesting.net/framework/unittest/unittest-introduction/
http://docs.python-guide.org/en/latest/writing/tests/
https://jeffknupp.com/blog/2013/12/09/improve-your-python-understanding-unit-testing/

Also, please don't create a separate file for each test.

hydra-redis/hydra_graph.py
```diff
@@ -0,0 +1,80 @@
...
+def get_apistructure(entrypoint_obj, url_node, api_doc):
+    """ It breaks the endpoint into two parts collection and classes"""
+    collection_endpoint = {}
+    classes_endpoint = {}
+    collection = 0
+    classes = 0
+    print("split entrypoint into 2 types of endpoints collection and classes")
+    for obj in url_node.properties:
+        if obj != "@id" and obj != "@type" and obj != "@context":
+            if obj in api_doc.collections.keys():
```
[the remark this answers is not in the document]
sandeepsajan0: I am not extracting the supportedProperty here I just extract the name and id of endpoints and divide into 2 types in this way:
`api_doc.entrypoint.get()` it will provide only and exact information about entrypoint that which endpoints are contained by entrypoint and with the use of `api_doc.collections.keys()` I can differentiate in class_endpoints and collection_endpoints.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def endpointclasses(self, classes_endpoint, url_node, api_doc, base_url):
+        """Node for every class which have an endpoint."""
+        print("classes endpoint or accessing classes")
+
+        for obj in classes_endpoint:
+            obj_method = []
+            obj_prop = []
+            properties = []
+            obj_properties = {}
```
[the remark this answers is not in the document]
sandeepsajan0: I'll provide a comment line for these types of naming conventions. It would be fine?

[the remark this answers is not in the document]
xadahiya: **NO!** You should always use proper naming conventions.

hydra-redis/hydra_graph.py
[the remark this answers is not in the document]
xadahiya: I know but it would be better if you use api_doc everywhere. Also, we've already separated Collection endpoints and Class endpoints using doc_writer why bother doing that again?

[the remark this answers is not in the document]
xadahiya: Also, this will be a much cleaner approach.

hydra-redis/classes_objects.py
```diff
@@ -0,0 +1,118 @@
...
+            member = {}
+            print("check endpoint url....loading data")
+            new_url = base_url + classes_endpoint[obj]
+            # url for the classes endpoint
+            print(new_url)
+            response = ur.urlopen(new_url)
+            new_file = json.loads(response.read().decode('utf-8'))
```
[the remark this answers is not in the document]
sandeepsajan0: I have checked already about its endpoint. So, I confirm that it is an endpoint. I am storing the fetched data in the endpoint node property So, I think it is beneficial to fetch data before creating the node because endpoint node contains all its data in its property section.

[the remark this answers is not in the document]
sandeepsajan0: Yeah, for that type of problem I have a check [here](https://github.com/HTTP-APIs/python-hydra-agent/pull/13/files#diff-48a117953692e711931c1871db9e9333R94) and I'll also add a check there is `if support_property.title in new_file:` I think It is working for that issue after adding this because I have checked it locally. I'll update it early.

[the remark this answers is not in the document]
xadahiya: The way you've implemented it is probably not the best way to do this. Please change it.

hydra-redis/classes_objects.py
sandeepsajan0: here we are storing those properties which are an object itself And that's type of property has title like `vocab:title`.

xadahiya: Yes, I know. That's what I said in the comment above. Please read it carefully.

hydra-redis/test/redis_test.py
```diff
@@ -0,0 +1,103 @@
...
+                "p.properties" in property_list and
+                "p.type" in property_list):
+            return True
+        else:
+            return False
+
+
+class TestRedisStructure(unittest.TestCase):
+    test_redis = Tests()
+
```
Mec-iS: Use:
```
@classmethod
def setUpClass(self):
    ...
```
To create a test database and dump some bulk data in it.

hydra-redis/test/redis_test.py
```diff
@@ -0,0 +1,103 @@
...
+    def test_entrypoint(self):
+        self.assertTrue(self.test_redis.entry_point())
+
+    def test_collectionEndpoints(self):
+        self.assertTrue(self.test_redis.collection_endpoints())
+
+    def test_classEndpoints(self):
+        self.assertTrue(self.test_redis.class_endpoints())
+
```
Mec-iS: Use:
```
@classmethod
def tearDownClass(cls):
    ...
```
to drop the test database. You can create a database with number `5` in Redis as test database. Adn keep `0` as the main database.

sandeepsajan0: Here as you say, I think we want to test that Redis working fine or not. For this, we have to create a database number `5` in Redis as test database and add some data to it like `set key value` and then check it like `get key` and after checking we have to drop the test database. And in this whole process, we will not touch the main database `0` it should remain same. 
Am I right? or we have to use database`0` for dumping data in database`5`?

xadahiya: No, you don't need to use database 0

xadahiya:
selection_042

@sandeepsajan0 getting this error. Also, add `graphviz` to requirements and also add some instructions  for `graphviz` executable installation.

Also, create a wiki explaining how your code works. You can use the project wiki https://github.com/HTTP-APIs/python-hydra-agent/wiki

xadahiya:
One more thing, what about the Google server you were about to setup?

sandeepsajan0:
@xadahiya  the error you have got, I think it is due to the server. You can use the server that I had set up is here https://storage.googleapis.com/api2/api (also in code as a comment)  the code is working fine with that. And yeah I was waiting for your review on the code I'll also write a blog as a documentation of code and graph. Yeah I'll update the requirements as add graphviz.

xadahiya:
Okay, cool! It's working nice work. The error with the other server is due to 404. Aren't you handling 404 errors?

xadahiya:
Also, write a blog post as well as the wiki here. Please make sure you write a bit longer and more explanatory blog post this time.

sandeepsajan0:
No, I am not handling 404 error. Is there any need to handle with that? Should we use continue for 404 error?

xadahiya:
We'll discuss this after the evaluations. Leave it as it is for now.

xadahiya:
@sandeepsajan0 You need to add anything before I merge?

sandeepsajan0:
No, I have nothing to add yet.

hydra-redis/hydra_graph.py
```diff
@@ -0,0 +1,89 @@
...
+        alias="Entrypoint",
+        properties=entrypoint_properties)
+    redis_graph.add_node(entrypoint_node)
+    return get_apistructure(entrypoint_node, api_doc)
+
+
+if __name__ == "__main__":
+    redis_con = redis.Redis(host='localhost', port=6379)
+    redis_graph = Graph("apidoc", redis_con)
+    url = "https://storage.googleapis.com/api2/api"
```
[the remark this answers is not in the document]
sandeepsajan0: I'll see it later, For now, we can use the demo server  http://35.224.198.158:8080/api.
Actually, It was my first time for making API server. So, there can be any mistake by me.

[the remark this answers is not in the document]
xadahiya: Won't this server show 404 errors like before?

PR #23, opened 2018-06-16
Implement querying mechanism

sandeepsajan0:
Now it loads only the API structure from the API doc like classendpoints, collectionendpoints, entrypoints and some of these properties and operations, not loads the data from the endpoint. Please suggest any changes.


xadahiya:
@sandeepsajan0  The whole agent is supposed to work as a single thing. Why are you rewriting the code? You're supposed to refactor the code you wrote earlier to fit our needs not rewrite the entire thing.

sandeepsajan0:
Now, only the graph loads in Redis but after the first query data will be load in Redis in a way like for collection_endpoints, we will give `endpoint_id` to `load_from_sever` function and it will load the data for this particular endpoint to the Redis and all based on user's query. The code may be updated after the querying mechanism.

hydra-redis/classes_objects.py
```diff
@@ -66,81 +66,98 @@ def objects_property(
             # set edge between the object and its parent object
             if endpoint_prop:
                 self.objects_property(
-                    object_node, endpoint_prop, entrypoint_node, api_doc)
+                    object_node, endpoint_prop, api_doc)
+
+    def load_from_server(
+            self,
+            class_endpoint_id,
```
xadahiya: All you need to fetch data from server is `url` or the endpoint. Refactor the code and create simple functions

sandeepsajan0:
The latest commit I have done for the update of querying mechanism and I am still working on it. So, review it later when I will update it next time. Now, I am working on the modification of graph and also on querying mechanism. So, I'll update it soon.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+import hydra_graph as graph
+from hydrus.hydraspec import doc_maker
+import hydrus
+from collections_endpoint import CollectionEndpoints
+import urllib.request
+import json
+
+def final_file(url):
+    """ Open the given url and read and load the Json data."""
+    response = urllib.request.urlopen(url)
```
Mec-iS: `try-except` block needed here. Every I/O call has to be safely written into an exception block, same for the JSON serialization below.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+    return json.loads(response.read().decode('utf-8'))
+
+def split_data(count,objects):
+    for obj in objects:
+        string = obj.decode('utf-8')
+        map_string = map(str.strip,string.split(','))
+        property_list = list(map_string)
+        check = property_list.pop()
+        property_list.append(check.replace("\x00",""))
```
Mec-iS: How it happens that this character `"\x00"` ends up here?

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+        print (property_list)
+
+def endpoints_query(query):
+    r = redis.StrictRedis(host='localhost', port=6379, db=0)
```
Mec-iS: You should factorized this at application run-time, and make the connection available through a proxy class or function like:
```
class RedisProxy:
    def __init__(self):
        self.connection = redis.StrictRedis...
    def get_connection(self):
        return self.connectoin
```

Mec-iS:
See annotations in the code.

Other annotations:
* I don't see any comment explaining the procedures
* try to apply basic patterns like Proxy or Facade, I know it seems hard in the beginning but you should try to give a better design to the codebase since the beginning, refactoring has to speed up, you can write better scripts since the beginning if the code is more ordered
* in general, work should go on faster, with more iterations

hydra-redis/querying_mechanism.py
sandeepsajan0: `"\x00" is null character and In C it works also as a string terminator character. [Here](https://stackoverflow.com/questions/24948139/why-get-string-from-redis-will-miss-blank-space) is something related to this.

sandeepsajan0:
Thanks @Mec-iS , Actually I am still working on it and I will update it soon with all these things you have mentioned.

hydra-redis/hydra_graph.py
```diff
@@ -17,7 +17,9 @@ def final_file(url):
 
 def get_apistructure(entrypoint_node, api_doc):
     """ It breaks the endpoint into two parts collection and classes"""
+    global collection_endpoints
```
xadahiya: Why are you using global variables here?

hydra-redis/hydra_graph.py
```diff
@@ -64,26 +66,32 @@ def get_endpoints(api_doc):
     redis_graph.add_node(entrypoint_node)
     return get_apistructure(entrypoint_node, api_doc)
 
-
-if __name__ == "__main__":
+def main(new_url,api_doc):
     redis_con = redis.Redis(host='localhost', port=6379)
+    global redis_graph
```
xadahiya: Instead of creating a global variable for the graph you should pass the graph as a function argument whenever required. **Global variables should be used very carefully. You should avoid using them unless it's absolutely necessary.**

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+        count = 0
+
+    if query == "collectionEndpoint":
+#        reply = r.execute_command('GRAPH.QUERY','apidoc','RETURN ',query,'')
+        reply  = r.execute_command('GRAPH.QUERY',
+                                   'apidoc','MATCH (p:collection) RETURN p')
+        count = 2
+
+    for objects in reply:
+        count+=1
```
xadahiya: What is count? How does it work?
Add DOCStrings and proper comments

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+
+def main():
+    initialize= 0
+    while(1):
+        if initialize ==0:
+            print("just initialize")
+            global url
+            url =input("url>>>")
+            apidoc = final_file(url + "/vocab")
+            global api_doc
```
xadahiya: Again, avoid using global variables

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,184 @@
...
+                r.execute_command('GRAPH.QUERY',
+                                  'apidoc', "MATCH (p:collection) RETURN p")
+        count = 0
+
+    if query == "classEndpoint":
+        reply = r.execute_command('GRAPH.QUERY',
+                                  'apidoc','MATCH (p:classes)' 'RETURN p')
+        count = 0
+
+    if query == "collectionEndpoint":
```
xadahiya: Try to write small and separate functions for different types of queries. It'll make the code more readable and easier to maintain.

sandeepsajan0:
Querying mechanism is working now but I am still working on some of its issues like:
* faceted index for every node properties.
* error handling.
* add more features.
* #24 
@Mec-iS  I have a point that it needs to indexing for every node's properties? If yes please elaborate a bit for me. As I think about it, it can partially resolve the #24 because if we use faceted indexing then we can show the data to the user with help of indexing. So, we don't need to access the graph for data and user will not get the problem of multiple data. Am I right? Please explain more.

hydra-redis/hydra_graph.py
sandeepsajan0: Actually, we need to import redis_graph attribute in querying_mechanism. So, I am using this as a global variable.

Mec-iS:
* Faceted indexes work because they index only few properties of an object. Indexing the full object means storing the object twice. For now we need only to index the unique id of an object, that's why I asked to work on a globally unique id for every entity.

Once we have unique ids, we can talk about synchronization and querying

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,463 @@
...
+
+def main():
+    """
+    Take URL as an input and make graph using initilize function.
+    Querying still user wants or still user enters the exit.
+    """
+    url = input("url>>>")
+    handle_data = HandleData()
+    apidoc = handle_data.load_data(url + "/vocab")
+    while (1):
```
xadahiya: Use `while(True)` instead. Also, you should display what errors occurred in APIDoc loading failed.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,463 @@
...
+            apidoc = handle_data.load_data(url + "/vocab")
+        else:
+            break
+
+    api_doc = doc_maker.create_doc(apidoc)
+    facades = QueryFacades(api_doc, url)
+    facades.initialize()
+    global check_list
+    check_list = []
+    while(1):
```
xadahiya: You should do something like [this](https://stackoverflow.com/questions/9340391/python-interactive-shell-type-application), it's much more elegant.

xadahiya: You can also create a `help()` to return the help strings.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,463 @@
...
+    print("for collection_endpoint:- show collectionEndpoint")
+    print("for members of collection_endpoint:-",
+          "show <collection_endpoint> members")
+    print("for properties of any member:-",
+          "show object<id_of_member> properties ")
+    print("for properties of objects:- show objects<endpoint> properties")
+    print("for collection properties:-",
+          "show <collection_endpoint> properties")
+    print("for classes properties:- show class<class_endpoint> properties")
+    print("for compare properties:-show <key> <value> and/or <key1> <value1>")
```
xadahiya: Clear this, use the method I've suggested above.

hydra-redis/classes_objects.py
```diff
@@ -56,91 +56,122 @@ def objects_property(
...
         if endpoint_property_list:
             for endpoint_property in endpoint_property_list:
                 for src_node in self.redis_graph.nodes.values():
                     if str(endpoint_property) == src_node.alias:
                         for endpoints in endpoint_property_list[
-                                               endpoint_property]:
+                                endpoint_property]:
                             for nodes in self.redis_graph.nodes.values():
                                 if endpoints == nodes.alias:
                                     self.addEdge(
```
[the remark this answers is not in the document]
sandeepsajan0: If we got the endpoint once in redis_graph nodes then no need to go ahead in the redis_graph. So, I am using the break for that.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,463 @@
...
+            if query.count(" ") > 1:
+                key, value, query = query.split(" ", 2)
+                faceted_list.append(self.faceted_key(key, value))
+            else:
+                key, value = query.split(" ")
+                query = ""
+                faceted_list.append(self.faceted_key(key, value))
+            if len(query) > 0:
+                operation, query = query.split(" ", 1)
+                if operation == "or":
```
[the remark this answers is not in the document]
sandeepsajan0: Yeah, there are some issues like this. I am working on it.

hydra-redis/querying_mechanism.py
sandeepsajan0: Okay.
Yeah, we are displaying the error in HandleData (line 26-30).

hydra-redis/hydra_graph.py
```diff
@@ -17,7 +17,9 @@ def final_file(url):
 
 def get_apistructure(entrypoint_node, api_doc):
```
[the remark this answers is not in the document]
xadahiya: Convert hydra_graph code into a class,  define `redis_graph` and other global variables you're using in the file and pass them as arguments when you create the instance for hydra_graph.

hydra-redis/querying_mechanism.py
```diff
@@ -0,0 +1,463 @@
...
+
+    def show_data(self, get_data):
+        """
+        Make the given data readable, because now it is in binary string form.
+        Count is using for avoid stuffs like query internal execution time.
+        """
+        count = 0
+        for objects in get_data:
+            count += 1
+            if count % 2 != 0:
```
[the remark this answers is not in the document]
xadahiya: Still not very clear, please add some example

hydra-redis/test/querying_test.py
```diff
@@ -0,0 +1,127 @@
+import unittest
+import urllib.request
+import json
+from hydrus.hydraspec import doc_maker
+from os import sys, path
+
+PARENT_DIR = path.dirname(path.dirname(path.abspath(__file__)))
+sys.path.append(PARENT_DIR)
+import querying_mechanism
```
xadahiya: Use absolute imports everywhere not relative imports

hydra-redis/test/querying_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+class Tests:
+
+    def endpointTest(self):
+        """Test for all the endpoints class + collection endpoints"""
+        query = "show endpoints"
+        data = query_facades.user_query(query)
+        if len(data) == 11:
```
xadahiya: You need to use assert statements in tests. Also, don't just compare the length you need to test data as well.

hydra-redis/test/querying_test.py
```diff
@@ -0,0 +1,127 @@
...
+        # Here are find the datastream only for those which have DroneID 2.
+        query = "show DroneID 2 and type Datastream"
+        data = query_facades.user_query(query)
+        if data == check_data:
+            return True
+        else:
+            return False
+
+
+class TestQueryingMechanism(unittest.TestCase):
```
xadahiya: You should define tests here only. Why are you defining them outside?

hydra-redis/test/querying_test.py
```diff
@@ -0,0 +1,127 @@
...
+    query_test = Tests()
```
xadahiya: Also, use teardown and setup methods for testing.

hydra-redis/test/querying_test.py
sandeepsajan0: I think we don't need to test for all the endpoint because we have two different test for class endpoints and collection endpoints. So, These two can cover this endpoint test. And Yeah, I have change comparison of length to comparison of data. I'll push it early.

hydra-redis/test/querying_test.py
sandeepsajan0: I am trouble here, I have tried or read many things but I don't find any for that. Please help for that.

hydra-redis/test/querying_test.py
sandeepsajan0: I can not figure out, what should I put in setUp and tearDown function? I think there are the different things in every test because of the different query in every test nothing is similar to put in setUp function.

sandeepsajan0:
I think, only one major issue is remaining that user can input `and` and `or` both in one query.
I need some time for that. Can I open an issue for that and can discuss there if required?

hydra-redis/test/querying_test.py
```diff
@@ -76,44 +81,20 @@ def ControllerLogCollectionmemberTest(self):
...
-
-    def test_5_ControllerLogCollectionmember(self):
-        self.assertTrue(self.query_test.ControllerLogCollectionmemberTest())
-
-    def test_6_DatastreamCollectionmember(self):
-        self.assertTrue(self.query_test.DatastreamCollectionmemberTest())
+        self.assertEqual(data,check_data)
 
 
 if __name__ == "__main__":
```
xadahiya: Sandeep you're supposed to initialize the graph and everything in the `setUp` function. It is meant to do the same. Rest everything looks fine for now. Create an issue for the `and or` query part then we can merge this.

sandeepsajan0: @xadahiya  If we initialize the graph every time for every test with `setup` method then there will be data stored multiple times in Redis server(because Redis server is running independently to the files). And It'll provide duplicate data for which we already have an issue #24 ( I am working on it). And due to duplicate data test cases will fail.

hydra_redis/test/querying_test.py
```diff
@@ -0,0 +1,99 @@
...
+        check_data = ['/api/DatastreamCollection/19']
+        query = "show DatastreamCollection members"
+        data = query_facades.user_query(query)
+        # Here are find the datastream only for those which have DroneID 2.
+        query = "show DroneID 2 and type Datastream"
+        data = query_facades.user_query(query)
+        self.assertEqual(data,check_data)
+
+
+if __name__ == "__main__":
```
xadahiya: @sandeepsajan0  You need to include all this in the `setUp` function, rest everything looks fine for now. Fix this then we can merge your PR.

xadahiya: Just call `unittest.main()` here

Issue #24, opened 2018-06-23
Add new data in Redis memory


sandeepsajan0:
Every time when I am trying to add some more data to Redis memory or update the Redis graph, I get the unwanted result like double of previous data means `previous_data + (previous_data + new_data)` but I want to add only new data.

Some different solutions suggested by @Mec-iS 

 * Build a secondary index using a Redis Set

 *  use hashing on objects, and store only the first x characters

 * Some Redis commands have the "NX" suffix to ensure that a key or a member is created only if does
 not exist
 
and 

    In general, this is an "upsert" problem that can be solved with transactions (see MULTI EXEC and    WATCH commands in Redis)`

sandeepsajan0:
@Mec-iS  @xadahiya  As you suggest the things regarding the solution, I have read about it. But I can not understand how I can use these solutions to Redis graph.
Ex. of problem:
we add a node and commit it with redis_graph(connection).
` node1 = Node( "collection", endpoint, node_properties)`
`redis_graph.add_Node(node)`
`redis_graph.commit()`
then redis_graph have an node with label ="collection" and alias = endpoint and properties as node_properties
Now, if we add 1 more node to the same graph(redis_graph) like this:
`node2 = Node( "objects", endpoint1, node_properties1)`
`redis_graph.add_node(node2)`
`redis_graph.commit()`
Now we expects that we would have 2 nodes in graph but when we access the node of graph there are three nodes 2 identical nodes for node1 and 1 node for node2.
It occurs because when we have committed the graph second time then it has two nodes node1 and node2 while node1 is already committed. So, node1 committed 2 times.
 
And I am unable to use some Redis command like `NX`  with Redis graph command like `redis_graph.commit().` So, please tell me how I can handle now.

sandeepsajan0:
fixed in #44

PR #44, opened 2018-07-17
update "and" and "or" combine query

### LINKED ISSUE #34: "AND" and "OR" combine query is not working
Now, queries can take only "and" or only "or" as operators but if a user tries to give both "and" and "or" operations in a query then the agent is not working.

Actually, we are using set operations `sinter` and `sunion` separately here. So, this issue raised.
Now we have to use them combined for complex queries.
And the Important thing here is to identify that where should be used `sinter` and `sunion`.
I am working on it.

sandeepsajan0:
Fixes #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,19 +14,19 @@ With the help of Redis, clients become faster and easier to query the data.
 
 **NOTE:** You'll need to use python3.
 
-To install or setup the client environment, you have to run:
-
-     python3 setup.py install
-
-or,
-
 To install only requirements:
    
     pip3 install -r requirements.txt
 
-To install Redis and start Redis server:
+or,
 
-    cd hydra_redis
+To install or setup the client environment, you have to run:
+
+     python3 setup.py install
+
+
+To install Redis and start Redis server(also if Redis has installed):
+
     ./redis_setup.sh
 
 ## Quickstart
@@ -96,19 +96,21 @@ The client takes the query as input, like:
 you can query as following querying formats:
 
 ```
-    print("for endpoint:- show endpoint")
-    print("for class_endpoint:- show classEndpoint")
-    print("for collection_endpoint:- show collectionEndpoint")
-    print("for members of collection_endpoint:-",
+    print("querying format")
+    print("Get all endpoints:- show endpoints")
+    print("Get all class_endpoints:- show classEndpoints")
+    print("Get all collection_endpoints:- show collectionEndpoints")
+    print("Get all members of collection_endpoint:-",
           "show <collection_endpoint> members")
-    print("for properties of any member:-",
+    print("Get all properties of objects:-",
+          "show objects<endpoint_type> properties")
+    print("Get all properties of any member:-",
           "show object<id_of_member> properties ")
-    print("for properties of objects:-show objects<endpoint_type> properties")
-    print("for collection properties:-",
-          "show <collection_endpoint> properties")
-    print("for classes properties:- show class<class_endpoint> properties")
-    print("for compare properties:-show <key> <value> and/or <key1> <value1>")
-    print("for using both opeartions(and,or) you should use brackets like:-",
+    print("Get all classes properties:-show class<class_endpoint> properties")
+    print("Get data with compare properties:-",
+          "show <key> <value> and/or <key1> <value1>")
+    print("Get data by using both opeartions(and,or)",
+          " you should use brackets like:-",
           "show model xyz and (name Drone1 or name Drone2)",
           "or, show <key> <value> and (<key> <value> or <key> <value>)")
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ you can query as following querying formats:
           "show <collection_endpoint> properties")
     print("for classes properties:- show class<class_endpoint> properties")
     print("for compare properties:-show <key> <value> and/or <key1> <value1>")
+    print("for using both opeartions(and,or) you should use brackets like:-",
+          "show model xyz and (name Drone1 or name Drone2)",
+          "or, show <key> <value> and (<key> <value> or <key> <value>)")
+
 ```
 
 Query test can be done like this:
--- a/hydra_redis/querying_mechanism.py
+++ b/hydra_redis/querying_mechanism.py
@@ -1,4 +1,3 @@
-import redis
 import random
 import string
 from hydra_redis.hydra_graph import InitialGraph
@@ -50,7 +49,7 @@ class HandleData:
             # Show data only for odd value of count.
             # because for even value it contains stuffs like time and etc.
             # ex: Redis provide data like if we query class endpoint
-            # output like: 
+            # output like:
             # [[endpoints in byte object form],[query execution time:0.5ms]]
             # So with the help of count, byte object convert to string
             # and also show only useful strings not the query execution time.
@@ -65,7 +64,6 @@ class HandleData:
                         #                        print(property_list)
                         all_property_lists.append(property_list)
         return all_property_lists
-
 
 
 class EndpointQuery:
@@ -131,7 +129,7 @@ class CollectionmembersQuery:
     Check_list is using for track which collection endpoint data is in Redis.
     """
 
-    def __init__(self, api_doc, url,graph):
+    def __init__(self, api_doc, url, graph):
         self.redis_connection = RedisProxy()
         self.handle_data = HandleData()
         self.connection = self.redis_connection.get_connection()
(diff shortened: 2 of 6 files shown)


xadahiya:
image
@sandeepsajan0 Getting this error. Also, you should format the output in such a way that it can be easily understood.

hydra_redis/querying_mechanism.py
```diff
@@ -526,6 +608,9 @@ def help():
           "show <collection_endpoint> properties")
     print("for classes properties:- show class<class_endpoint> properties")
     print("for compare properties:-show <key> <value> and/or <key1> <value1>")
+    print("for using both opeartions(and,or) you should use brackets like:-",
```
xadahiya: Update the commands like this - 
```
Querying format
Get all endpoints :- show endpoints
Get all Class endpoints :- show classEndpoints
Get all Collection endpoints :- show collectionEndpoints
Get all members of a collection endpoint :- show <collection_endpoint> members
Get all properties of an endpoint member :- show object <id_of_member> properties
Get all properties of a class endpoint :- show object <class_endpoint_type> properties
```

Update other commands according to this format as well.

hydra_redis/querying_mechanism.py
```diff
@@ -526,6 +608,9 @@ def help():
           "show <collection_endpoint> properties")
```
xadahiya: How is this different from command at line 602?

hydra_redis/querying_mechanism.py
```diff
@@ -526,6 +608,9 @@ def help():
...
     print("for classes properties:- show class<class_endpoint> properties")
```
xadahiya: Same goes for this as well ( line 606)

README.md
```diff
@@ -108,6 +108,10 @@ you can query as following querying formats:
           "show <collection_endpoint> properties")
     print("for classes properties:- show class<class_endpoint> properties")
     print("for compare properties:-show <key> <value> and/or <key1> <value1>")
+    print("for using both opeartions(and,or) you should use brackets like:-",
+          "show model xyz and (name Drone1 or name Drone2)",
+          "or, show <key> <value> and (<key> <value> or <key> <value>)")
```
xadahiya: Update README according the querying format I proposed.

xadahiya:
Update the redis script to check if docker and redis are already configured. If yes, then simply start redis or setup redis and docker.

Also, there should be a command in README to run the redis docker instance if user has already set everything up using the redis setup script.

hydra_redis/querying_mechanism.py
```diff
@@ -354,6 +354,69 @@ def faceted_key(self, key, value):
...
+        It takes the arguement as list(query_list)
+        which contains the faceted indexes and operation and brackets also.
+        List Ex:['fs:model:xyz', 'and', '(', 'fs:name:Drone1', 'or',
+                'fs:name:Drone2', ')']
+                for query "model xyz and (name Drone1 or name Drone2)"
+        :param query_list: get a list of faceted indexes and operations
+        :param return: get data from the Redis memory for specific query.
+        """
+        # check if there is both "and" and "or" with help of bracket.
+        if ")" not in query_list:
```
xadahiya: What about the precedence thing we talked about?

sandeepsajan0:
The error that you are facing is due to the wrong query, it should be `collectionEndpoints` instead of `collectionEndpoint`. Yeah, It can be an issue that client should give warning and etc but not exit with showing error(error handling). I will make an issue for that.

hydra_redis/querying_mechanism.py
sandeepsajan0: It is different internally if `members` is query then it will check Are these endpoints members loaded or not in Redis?
but if `properties` is a query then the client doesn't check the data it will directly access data from Redis.

hydra_redis/querying_mechanism.py
sandeepsajan0: In line 606, the client shows properties of members of a collection(objects) but in line 609, it shows properties of class endpoints.

hydra_redis/querying_mechanism.py
sandeepsajan0: okay, I'll update it.

hydra_redis/querying_mechanism.py
sandeepsajan0: I'll work on that. Now I am trying to make Dockerfile. And I think brackets are good to use because it makes query simpler and easily understandable.
Because using brackets a simple query can asked in many diff. ways like:
`a and b and c or d or e and f`
it can be like these:
`(a and b and c) or d or (e and f)` and
`a and b and (c or d or e) and f` and 
`a and b and ((c or d or e) and f`

What do you think?

hydra_redis/querying_mechanism.py
xadahiya: Can you provide an example?

hydra_redis/querying_mechanism.py
xadahiya: What properties of class endpoints? Can you provide an example?

hydra_redis/querying_mechanism.py
xadahiya: Yeah, I agree. There's no question the format with brackets is better and more understandable but supporting queries without brackets would also be cool. It's just a add-on sort of thing to support a wider format of queries.

xadahiya:
Then why are you showing `collectionEndpoint` in help()?

xadahiya:
Also, if some query is incorrect the agent should not stop, it should prompt incorrect query with some message.

xadahiya:
Duplicate data is being saved in Redis if user enters some incorrect query and the agent is restarted again. You should add data only if it doesn't exist in redis and replace data only in case of some data change at server.

xadahiya:
Ideally, we should be able to run multiple agent instances (in different terminals) using the same redis backend.

xadahiya:
Also, what's the progress on keeping data on the agent's side in sync with server (The use of unique IDs)?

xadahiya:
Also, @sandeepsajan0 move the redis setup script to parent directory.

sandeepsajan0:
>Then why are you showing collectionEndpoint in help()?

sorry, typo I'll update it.

Regarding duplicate data, we have an issue and I'll work on that after it.
And Regarding multiple instances, it is working but we have to change something in code because in multiple instances case, it is storing duplicate data.

And for the use of unique id, the client should get the data from the server with unique id then we can update some part of code according to requirements.

hydra_redis/querying_mechanism.py
sandeepsajan0: class properties are like: 'Location' for class Location.
But objects properties are like: ['State'", "'name'", "'model'", "'MaxSpeed'", "'Sensor'] for DroneCollection/1 object.

hydra_redis/querying_mechanism.py
sandeepsajan0: Yeah, I think it is same as line 606. It should be removed from here. I'll update it.

hydra_redis/querying_mechanism.py
sandeepsajan0: Okay, Can I update it later? Because I think it is not a high priority issue like others.
After some other issues, I'll update it.

xadahiya:
@sandeepsajan0 How much time do you need to fix all the issues I mentioned. We don't have much time left.

xadahiya:
@sandeepsajan0  please comment explaining the restart issue fix here.

sandeepsajan0:
Now, If the agent will restart without restarting of Redis server then the agent will check that data for given URL and endpoint is stored in Redis or not. If data will be stored in Redis, it will not fetch data again. 
For this purpose, I have added the faceted index like `fs:url` for checking the URLs and `fs:endpoints` for checking the endpoints. Both indexes are a type of `set` which will store only different url and endpoints. The agent will check every time that url or endpoint is in `set` or not. With the help of `fs:endpoints`, agent also able to access multiple instances without storing data for every instance.

xadahiya:
@sandeepsajan0 Any progress?

sandeepsajan0:
@xadahiya  I am trying but I haven't got any idea with the help of which we can handle duplicate data.
I read and understand the redisgraph client.py file but I can't get any idea to change it with our requirements.
But I am still trying. And if you have any idea then please share I'll try that.

sandeepsajan0:
@xadahiya  I got an idea for Duplicate data. 
idea:  We can remove the old and not necessary data from Redis every time when we add new data to it because according to issue #24 . Old Data becomes double everytime when we save new data. So, if we erase old data then there will only new data it means whole data and no duplicate data.
I think it can work. I'll try to it.
what do you say?

sandeepsajan0:
Yeah, It is working fine now. I have implemented the above idea and delete only the old data which was added by redis_graph. It'll not touch the data which was not added by redis_graph or added directly in Redis. So, Now issue #24  solved. 
@xadahiya  please take a look into it.

xadahiya:
@sandeepsajan0 this a hack but if this works then I suppose it's fine for now. I'll try to figure out a better way to resolve this issue over the weekend.

xadahiya:
Moved to  #48

PR #45, opened 2018-07-20
update redis-setup.sh

sandeepsajan0:
Now if Redis is installed, then it will only run the server not install Redis and docker every time.

### DIFF
--- a/redis_setup.sh
+++ b/redis_setup.sh
@@ -1,5 +1,5 @@
 
-#It will remove all the previous docker engine
+#It will check, if docker is not installed then install it.
 docker -v
 if [ echo "$?" = "127" ]
 then
@@ -22,7 +22,7 @@ else
     echo "Docker is already installed"
 fi
 
-# after getting the docker-ce
+# after getting the docker-ce, check if `redislabs/redisgraph` docker image is not installed then install ii. 
 if [ -z "$(docker images -q redislabs/redisgraph:latest)" ]
 then
     echo "Docker already have a redislabs/redisgraph image"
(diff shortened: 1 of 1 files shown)


redis_setup.sh
```diff
@@ -15,10 +18,19 @@ sudo add-apt-repository    "deb [arch=amd64] https://download.docker.com/linux/u
 sudo apt-get update
 sudo apt-get install docker-ce
 
+else
+    echo "yesss exist"
```
xadahiya: Fix these messages. Please use something professional.

sandeepsajan0: ohh sorry! I was using it for checking.

redis_setup.sh
```diff
@@ -1,5 +1,8 @@
 
 #It will remove all the previous docker engine
+docker -v
+if [ echo "$?" = "127" ]
```
xadahiya: What is this?

redis_setup.sh
xadahiya:
```
./redis_setup.sh 
Docker version 18.06.0-ce, build 0ffa825
./redis_setup.sh: line 4: [: too many arguments
Docker is already installed
``` 
Fix this error message

redis_setup.sh
sandeepsajan0: if there will be no docker in machine, the command `docker -v` will throw error `127`. But if docker is already present, the command will throw  `0`. 
So, it is just a check for docker. And same I have done in line 26, for check `redislabs/redisgraph` docker image.

xadahiya: > #It will remove all the previous docker engine
Update this comment to explain this then we can merge this.

PR #48, opened 2018-07-26
Fix AND and OR query

xadahiya:


sandeepsajan0:
You haven't change all the file that has changed in #44 (like hydra_graph.py) It is not working fine now.
While #44  was working fine. there was only a dockerfile problem.
And I also don't understand why you have remove the code that was using for handle duplicate data.
And for docker error, I'll try with another system once because I can't figure out what is problem in my local packaging. After test on another system I'll tell you about that , hopefully soon.

xadahiya:
Sandeep it's your exact same pr with my one additional commit. Try to run it first.

xadahiya:
Test on your system only, it'll work fine now

sandeepsajan0:
Actually, when I tried to run both `db` for Redis and `client` for querying_mechanism combined with the use of docker compose then it gives the error to me but when I tried `db` for Redis(same docker image) and run querying_mechanism manually then it works fine.

PR #51, opened 2018-08-03
update error handle

sandeepsajan0:
@xadahiya  Please let me know if you get any breakdown still now.

### DIFF
--- a/hydra_redis/classes_objects.py
+++ b/hydra_redis/classes_objects.py
@@ -119,8 +119,19 @@ class ClassEndpoints:
         # new_url is url for the classes endpoint
         new_url = base_url + "/" + endpoint
         # retreiving data for the classes endpoint from server
-        response = urllib.request.urlopen(new_url)
-        new_file = json.loads(response.read().decode('utf-8'))
+        try:
+            response = urllib.request.urlopen(url)
+        except HTTPError as e:
+            print('Error code: ', e.code)
+            return ("error")
+        except URLError as e:
+            print('Reason: ', e.reason)
+            return ("error")
+        except ValueError as e:
+            print("value error:",e)
+            return ("error")
+        else:
+            new_file = json.loads(response.read().decode('utf-8'))
         # endpoint_property store all properties which is class/object but not
         # endpoint.
         for support_property in api_doc.parsed_classes[
--- a/hydra_redis/classes_objects.py
+++ b/hydra_redis/classes_objects.py
@@ -1,7 +1,15 @@
 import urllib.request
 import json
+import logging
 from redisgraph import Node, Edge
+from urllib.error import URLError, HTTPError
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+class RequestError(Exception):
+    """A class for client-side exceptions"""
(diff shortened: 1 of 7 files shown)


hydra_redis/collections_endpoint.py
```diff
@@ -16,8 +17,19 @@ def fetch_data(self, new_url):
         :param new_url: url for fetching the data.
         :return: loaded data.
         """
-        response = urllib.request.urlopen(new_url)
-        return json.loads(response.read().decode('utf-8'))
+        try:
+            response = urllib.request.urlopen(new_url)
+        except HTTPError as e:
+            print('Error code: ', e.code)
```
Mec-iS: These prints should be using a logger from `logging` library: https://docs.python.org/3/howto/logging-cookbook.html

sandeepsajan0:
No, we have tests only for few functions. Actually, we have dynamic data in the server. So, it was not easy to test them with unpredictable data. And, also for other tests, I have made a google cloud bucket but I think, now this link is broken due to some billing problem(I guess). I'll make another bucket with the same data and then update the tests link(possibly today).

hydra_redis/classes_objects.py
```diff
@@ -1,7 +1,15 @@
 import urllib.request
 import json
+import logging
 from redisgraph import Node, Edge
+from urllib.error import URLError, HTTPError
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+class RequestError(Exception):
```
xadahiya: @sandeepsajan0 What is this doing?

xadahiya: No, it's fine.

xadahiya:
@sandeepsajan0 Fix hydrus in requirements.txt

hydra_redis/classes_objects.py
sandeepsajan0: Can we merge it?

xadahiya: annotation

Getting this error @sandeepsajan0

sandeepsajan0: Are you trying the update code? Because It is running fine with me.
no_error

And also I can try once run the setup.py file.

xadahiya: It's probably a redisgraph issue. I think they've changed the structure a bit. Delete all docker imager and containers, create a new virtualenv and try again. You should see this error.

sandeepsajan0: Yeah, I got it!

Mec-iS:
@sandeepsajan0 
>  It can be tested by running the different queries(As I have tested it).

You should add tests for the functions in the client and use `mocking` to create dummy responses so to avoid your tests depending from other parts of your system: https://docs.python.org/3.5/library/unittest.mock.html

If you need a response from a query to test your function, you should mock it so to test the function iteself and avoiding testing the other part of the system (that should be tested separately with its own tests).

hydra_redis/classes_objects.py
sandeepsajan0: fixed

xadahiya:
merging #51

PR #62, opened 2018-12-08
fix docker

PrajwalM2212:
Fix docker and setup.py


xadahiya:
@PrajwalM2212 why exactly are we changing from `hydra_redis` to `hydra_agent`?

PR #63, opened 2018-12-20
Add .travis.yml

xadahiya:


sandeepsajan0:
I think the error is due to `-e` because we already have `-e` attached to hydrus source in requirements.txt.

PR #69, opened 2019-01-11
Change to allow git repo package install

de-sh:
Maybe tests are failing due to them being in develop mode which doesn't load easily from git sources. This is a trial. *Close if unsatisfactory*.


de-sh:
I made this pull to further understand the why the tests were failing.

PR #73, opened 2019-01-22
Convert dict to string

### LINKED ISSUE #72: Error when running python hydra agent
screen shot 2019-01-22 at 10 50 18 pm

Hydra connection expects a` string` as a value but a `dictionary` is passed instead.

PrajwalM2212:
fixes #72 
`redis_connection.set` accepts a string as its value , but a dictionary was passed.  This has been solved by using json.dumps.

### DIFF
--- a/hydra_agent/hydra_graph.py
+++ b/hydra_agent/hydra_graph.py
@@ -55,7 +55,7 @@ class InitialGraph:
             alias="Entrypoint",
             properties=entrypoint_properties)
         self.redis_graph.add_node(entrypoint_node)
-        redis_connection.set("EntryPoint", entrypoint_properties)
+        redis_connection.set("EntryPoint", json.dumps(entrypoint_properties))
         return self.get_apistructure(entrypoint_node, api_doc)
 
 


Mec-iS:
not passing CI

PrajwalM2212:
@Mec-iS One of the reason for travis failing was this. But to make the CI pass we have to remove the dependency on  https://storage.googleapis.com/api4/api.

xadahiya:
@sandeepsajan0 Please review

hydra_agent/hydra_graph.py
```diff
@@ -55,7 +55,7 @@ def get_endpoints(self,api_doc, redis_connection):
             alias="Entrypoint",
             properties=entrypoint_properties)
         self.redis_graph.add_node(entrypoint_node)
-        redis_connection.set("EntryPoint", entrypoint_properties)
+        redis_connection.set("EntryPoint", json.dumps(entrypoint_properties))
```
sandeepsajan0: @PrajwalM2212  I think, we don't need to have that line of code because we already have a node for the same. So, Can you try to run without this line? And don't forget to send the screenshot.

Mec-iS:
waiting for feedback from @PrajwalM2212

ShravanDoda:
What' the status here? Without removing that line, `python-hydra-agent` wouldn't run.

sandeepsajan0:
fixed in #94

PR #75, opened 2019-01-22
Fix error occurring when url  or query is correct

### LINKED ISSUE #74: Error when correct url or query is entered
screen shot 2019-01-22 at 10 51 02 pm

Even when correct url is entered python hydra agent gives an error.

PrajwalM2212:
fixes #74 

Even though correct url is used even an addition of extra space would cause an error. This has been fixed by using `strip`

### DIFF
--- a/hydra_agent/querying_mechanism.py
+++ b/hydra_agent/querying_mechanism.py
@@ -658,7 +658,7 @@ def query(apidoc, url):
 
     while True:
         print("press exit to quit")
-        query = input(">>>")
+        query = input(">>>").strip()
         if query == "exit":
             break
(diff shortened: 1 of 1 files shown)


ShravanDoda:
Apart from whitespaces, error is also thrown when there's a '/' at the end of the url because then it sends a url request to //vocab instead of /vocab. It'd be good if you could create another commit for checking the presence of '/' at the end of the input

sandeepsajan0:
@PrajwalM2212  Please see the [comment](https://github.com/HTTP-APIs/python-hydra-agent/issues/74#issuecomment-460452032). It needs some more changes.

Issue #76, opened 2019-01-27
Logger format is incorrect


ShravanDoda:
The logger function in `querying_mechanism.py` isn't implemented correctly.

PR #77, opened 2019-01-27
Corrected the logger.info() method

ShravanDoda:
#76
Changed the format of code from `logger.info('Error Code: ', e.code)` to 
`logger.info('Error Code:{}'.format(e.code))`.
The previous format is not supported by logger. Due to the error,  a lot of exceptions are made when some wrong command is typed in the client while there should only be one exception.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "venv/bin/python"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "venv/bin/python"
-}
--- a/hydra_agent/querying_mechanism.py
+++ b/hydra_agent/querying_mechanism.py
@@ -31,13 +31,13 @@ class HandleData:
         try:
             response = urllib.request.urlopen(url)
         except HTTPError as e:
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
PrajwalM2212: You should remove this file from your PR.

PrajwalM2212:
@ShravanDoda It is good practice to create a new branch like `ShravanDoda:fix_logger` rather than directly sending a PR from `ShravanDoda:develop`.

.vscode/settings.json
PrajwalM2212: You can add this file to `.gitignore `

hydra_agent/querying_mechanism.py
```diff
@@ -31,13 +31,13 @@ def load_data(self, url):
         try:
             response = urllib.request.urlopen(url)
         except HTTPError as e:
-            logger.info('Error code: ', e.code)
+            logger.info('Error Code: {}'.format(e.code))
```
Mec-iS: Those should be `logger.error`

ShravanDoda:
@sandeepsajan0, the issues with url input and context manager have also been resolved in the above commit

PR #81, opened 2019-02-10
solve issue running redis_setup.sh

### LINKED ISSUE #80: Error when running redis_setup.sh in command line
when I run ./redis_setup.sh
it showed following error
"./redis_setup.sh: line 4: [: too many arguments
Docker is already installed" 
even when docker is not installed 
it is due to an error in argument of if-else in line 4

krritik:
fixes #80 

**Checklist**

- my branch is upto date with upstream/develop branch
- Everything works and tested for Python 3.5.2 and above

**Description**

 While running the shell file redis_setup.sh to download docker and other requirements. It gave an error and do not download the docker even when if docker was not installed. It was due to an error in argument of if-else statement . So if statement is never executed and always else is executed.

**Change logs**

fix error in if-else statement in redis_setup.sh

### DIFF
--- a/redis_setup.sh
+++ b/redis_setup.sh
@@ -1,7 +1,7 @@
 
 #It will check, if docker is not installed then install it.
 docker -v
-if [ echo "$?" = "127" ]
+if [ "$?" = "127" ]
 then
 sudo apt-get update
 sudo apt-get install docker


Mec-iS:
the repository shall use `docker-compose` to run. This script has to be removed.

PR #84, opened 2019-02-22
Improve algo

### LINKED ISSUE #83: Improve the algorithm used to connect endpoints

ShravanDoda:
Fixes #83 
This implementation works by creating a dictionary local to the `endpointclasses` method. The dictionary has aliases of the nodes as keys and the corresponding graph nodes as values. Then it uses another method `connect_nodes`  which takes in the dictionary as well as the `endpoint_property_list` as argument and adds edges between the nodes. This way it does not have to traverse through the whole graph (which contains `collectionEndpoints` as well while the dictionary only contains concerned `classEndpoints`). Although the space complexity of this method is greater than the current implementation but the time complexity is reduced (thanks to python's dictionary.get() method's average case time complexity of O(1)).

@Mec-iS @sandeepsajan0 kindly review.

### DIFF
--- a/hydra_agent/classes_objects.py
+++ b/hydra_agent/classes_objects.py
@@ -239,5 +239,4 @@ class ClassEndpoints:
             # set edge between the entrypoint and the class endpoint.
             self.addEdge(entrypoint_node, "has" + endpoint, class_object_node)
             node_alias[str(endpoint)] = class_object_node
-            endpoint_property_list[endpoint] = property_list
         self.connect_nodes(endpoint_property_list, node_alias)
--- a/hydra_agent/classes_objects.py
+++ b/hydra_agent/classes_objects.py
@@ -188,6 +188,13 @@ class ClassEndpoints:
         # save the new data.
         self.redis_graph.commit()
 
+    def connect_nodes(self, endpoint_properties, node_alias):
+        for class_endpoint in endpoint_properties:
+            src_node = node_alias.get(class_endpoint)
+            for endpoint in endpoint_properties[class_endpoint]:
+                node = node_alias.get(endpoint)
+                self.addEdge(src_node, 'has_endpoint_property', node)
+
     def endpointclasses(
             self,
             entrypoint_node,
@@ -200,6 +207,7 @@ class ClassEndpoints:
         """
         print("classes endpoint or accessing classes")
         endpoint_property_list = {}
+        node_alias = {}
         # endpoint_property_list contain all endpoints
         # which have other endpoints as a property ex: State.
         for endpoint in self.class_endpoints:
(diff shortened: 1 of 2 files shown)


Mec-iS:
Related to #72

hydra_agent/classes_objects.py
```diff
@@ -188,6 +188,13 @@ def load_from_server(
         # save the new data.
         self.redis_graph.commit()
 
+    def connect_nodes(self, endpoint_properties, node_alias):
+        for class_endpoint in endpoint_properties:
+            src_node = node_alias.get(class_endpoint)
+            for endpoint in endpoint_properties[class_endpoint]:
+                node = node_alias.get(endpoint)
```
sandeepsajan0: You are considering that `node_alias` have all nodes required for an edge. But for "classes property endpoint", the node can be anywhere in the graph. It is not necessary that it will be in class endpoints`node_alias`. So, We need to traverse the whole graph to find that property node.

ShravanDoda: @sandeepsajan0 all the nodes for class_endpoints are being created by a single call to `endpointclasses` method. As soon as each node is created, it's appended `node_alias`. The above implementation is only for adding edge between `entrypoint` node and `classEndpoint` (that's the only function that `endpointclasses` method serves). `object_property` method hasn't been changed

sandeepsajan0: line 232 is responsible for `Entrypoint` and `classEndpoint` connection. But the code you have changed is responsible for connecting the "classEndpoints nodes" to the "object_property nodes" which may be available already. See, `State` is an `object_property` for any "collection member" and also for "classEndpoint". So, It can happen that there will be a node of `object_property`. So, We don't need to create duplicate node else we have to search for that node and connect to that.

ShravanDoda: @sandeepsajan0 I think that the code I changed connected `classEndpoint`s to other `classEndpoint`s. If I have 2 `classEndpoint`s and one of them has a property whose value is equal to the other `classEndpoint`, the code should create an edge between them. Please correct me if I'm wrong. If this is so, then `node_alias` method should work

sandeepsajan0: Yes, @ShravanDoda  You are partially right. It may be 2 `classEndpoints` but it may also be a `property`. See the example, in hydrus/doc.json file there is a class `State` which is also a property of `Command` and other classes. But `State` can not get from `Entrypoint` endpoint.
So, It is not the same as to connect 2 classes node. And I guess there may still a small bug but not same as we are discussing. Try to read apidoc and then try to find or solve the bug. I will ping you if I'll get the bug.
This [documentation](https://medium.com/w3c-hydra-development-community/documentation-for-hydra-graph-cd9b2bd84884) may help you.

hydra_agent/hydra_graph.py
```diff
@@ -55,7 +55,6 @@ def get_endpoints(self,api_doc, redis_connection):
             alias="Entrypoint",
             properties=entrypoint_properties)
         self.redis_graph.add_node(entrypoint_node)
-        redis_connection.set("EntryPoint", entrypoint_properties)
```
sandeepsajan0: @ShravanDoda `classes_objects` and `collection-endpoints` also have the same error. 
I think @PrajwalM2212 also working on the same issue in PR #73. So, Please ask him.

hydra_agent/classes_objects.py
sandeepsajan0: @ShravanDoda As I can see, there is no bug because in `hydra_graph` module we first calling to `classes_enpoints` and then `collection_endpoints`. So, There cannot be a property which already present in the graph. So, we don't need to traverse the whole graph. I think, your code should work the same. I will check and merge it after merging #87.

hydra_agent/classes_objects.py
```diff
@@ -230,19 +238,6 @@ def endpointclasses(
...
-                            for nodes in self.redis_graph.nodes.values():
-                                if endpoints == nodes.alias:
-                                    self.addEdge(
-                                        src_node,
-                                        "has_endpoint_property",
-                                        nodes)
-                                    break
-                        break
+            node_alias[str(endpoint)] = class_object_node
+            endpoint_property_list[endpoint] = property_list
```
sandeepsajan0: Repetition of this line. see line 232.

ShravanDoda: @sandeepsajan0, I think I should close this PR (as @Mec-iS stated in #83)

sandeepsajan0: @ShravanDoda  You can close it. And try to do it in a better way as @Mec-iS suggest.

xadahiya:
Closing as requested

PR #85, opened 2019-02-25
Changes to url processing

### LINKED ISSUE #74: Error when correct url or query is entered

ShravanDoda:
Fixed #74 
Changes ->

- Added a `with` block to `urlopen` 

- Added a check for trailing slashes and used `urllib` tools to join url paths

Kindly review


ShravanDoda:
@sandeepsajan0 @Mec-iS the build is failing due to a `hydra_python_core` issue. We should integrate it before we proceed

Issue #86, opened 2019-02-25
Integrate python-hydra-agent and hydra-python-core


ShravanDoda:
`python-hydra-agent` has to be modified to support `hydra-python-core`. Travis CI build fails as well.

PR #87, opened 2019-02-25
Use 'hydra_python_core' module in place of 'hydrus.hydraspec'

de-sh:
**Solves** #86 

As has been done with moving out the hydra_python_core module that used to be hydraspec, the python hydra-agent needs to be updated similarly.

## Changes
- Reflected moving out of 'hydraspec' as hydra_python_core module, updated  all use of same within the repo.

### DIFF
--- a/hydra_agent/hydra_graph.py
+++ b/hydra_agent/hydra_graph.py
@@ -2,7 +2,7 @@ import redis
 from redisgraph import Graph, Node
 import urllib.request
 import json
-from hydra_python_core import doc_maker 
+from hydra_python_core import doc_maker, doc_writer
 from graphviz import Digraph
 from hydra_agent.classes_objects import ClassEndpoints,RequestError
 from hydra_agent.collections_endpoint import CollectionEndpoints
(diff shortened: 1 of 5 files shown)


hydra_agent/hydra_graph.py
```diff
@@ -21,12 +20,12 @@ def get_apistructure(self,entrypoint_node, api_doc):
         for support_property in api_doc.entrypoint.entrypoint.supportedProperty:
             if isinstance(
                     support_property,
-                    hydrus.hydraspec.doc_writer.EntryPointClass):
+                    doc_writer.EntryPointClass):
```
sandeepsajan0: @de-sh There is an error `doc_writer is not defined`. Fix this error.

ShravanDoda: Its not being imported at the top.

sandeepsajan0: yes @ShravanDoda, you are right.  @de-sh  import it.

de-sh: Done :+1:

ShravanDoda:
Travis config needs to change. It's still installing `hydrus`.

PR #89, opened 2019-03-02
Fixing Redis breaking change

Guttz:
Fixes # [86](https://github.com/HTTP-APIs/python-hydra-agent/issues/86)

### Checklist

### Description
Setting Redis version as 2 because of breaking change in Redis 3
#### Fixed

redis.exceptions.DataError: Invalid input of type: 'dict'. Convert to a byte, string or number first.
(redis_connection.set("EntryPoint", entrypoint_properties)...)

Set Redis version because of backwards incompatibility from Redis 2 to Redis 3, which now only accepts values as bytes, strings or numbers.

### DIFF
--- a/hydra_agent/querying_mechanism.py
+++ b/hydra_agent/querying_mechanism.py
@@ -4,7 +4,7 @@ import logging
 from hydra_agent.hydra_graph import InitialGraph
 import urllib.request
 import json
-from hydrus.hydraspec import doc_maker
+from hydra_python_core import doc_maker
 from urllib.error import URLError, HTTPError
 from hydra_agent.collections_endpoint import CollectionEndpoints
 from hydra_agent.classes_objects import ClassEndpoints,RequestError
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,7 @@ ENV/
 
 # mypy
 .mypy_cache/
+
+# pipenv generated filespip 
+Pipfile
+Pipfile.lock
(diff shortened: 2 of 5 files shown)


hydra_agent/querying_mechanism.py
```diff
@@ -4,7 +4,8 @@
 from hydra_agent.hydra_graph import InitialGraph
 import urllib.request
 import json
-from hydrus.hydraspec import doc_maker
+import hydra_python_core
+from hydra_python_core import doc_maker
```
xadahiya: Don't need both these imports. You can either import hydra_python_core and use hydra_python_core.doc_maker or import dic_maker directly. I don't see the reason to do both

de-sh:
Hey @Guttz, you may consider debasing to the latest commit merged into develop b952f19d93768c998e89e4f1a939af2152f82cfb

hydra_agent/querying_mechanism.py
Guttz: You are right. Already fixed and re-based according to @de-sh commit.

Guttz:
> Hey @Guttz, you may consider re-basing to the latest commit merged into develop [b952f19](https://github.com/HTTP-APIs/python-hydra-agent/commit/b952f19d93768c998e89e4f1a939af2152f82cfb)

Done @de-sh ! Re-based and proposed only the different changes.

PR #91, opened 2019-03-06
Added mock tests for querying

### LINKED ISSUE #88: Mocking is done improperly
## Current Behaviour:
Mocking is being done improperly.
```python
        check_data = [['p.properties', 'p.id', 'p.type'],
                      ["['Location']",'vocab:EntryPoint/Location','Location']]
        query = "show classEndpoints"
        EndpointQuery_get_classEndpoints = MagicMock(return_value=check_data)
        data = EndpointQuery_get_classEndpoints(query)
        print("testing classEndpoints...")
assert data == check_data
```
Correct me if I am missing something, but it seems like this test will always pass, same goes for all the tests. I can't understand why are we mocking Queries?
## Expected Behaviour:
As far as I know we are using moking so we can test the code without calling an actual (server) API, in that case I think we should only mock calls to the server(`url.get()`).

ShravanDoda:
Fixes #88

Mock tests for querying in `python-hydra-agent` have been added. @sandeepsajan0 kindly review

### DIFF
--- a/hydra_agent/tests/test_querying.py
+++ b/hydra_agent/tests/test_querying.py
@@ -5,18 +5,30 @@ from hydra_agent.querying_mechanism import EndpointQuery, CollectionmembersQuery
 
 
 class TestHandleData(unittest.TestCase):
+    """
+    TestCase for HandleData Class
+    """
     def setUp(self):
+        """Setting up HandleData object"""
         self.handle_data = HandleData()
 
     @patch('hydra_agent.querying_mechanism.json.loads', spec_set=True)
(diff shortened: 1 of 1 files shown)


Mec-iS:
This looks ok to me.

Please @xadahiya @sandeepsajan0 take a look

Mec-iS:
@ShravanDoda Please add docstrings to test cases

sandeepsajan0:
@ShravanDoda  Can you please provide a screenshot of passing test cases? Other things are looking fine.

Mec-iS:
Tests passed in Travis

PR #94, opened 2019-03-10
Removed the line which was explicity setting EntryPoint in redis

ShravanDoda:
Fixes #73 
This commit takes care of the removal of line which was setting EntryPoint in the redis


sandeepsajan0:
@ShravanDoda Also take a look on `collection_endpoints` line:130 and check it also.

Issue #101, opened 2019-03-25
[PRE-ALPHA] Implement all HTTP method on all endpoints


Mec-iS:
~Move on from the GET-only implementation of the PoC and implement a full client~

For more extensive explanation, go to: https://github.com/HTTP-APIs/hydra-python-agent/issues/101#issuecomment-497685202

invinciblycool:
Currently, the GET requests are made when using the `show` option in the agent. Do we want something similar? Probably implementing a `insert/delete/update` option in the agent.

Mec-iS:
@invinciblycool see #99

invinciblycool:
What I understand is, that the agent is currently capable of only fetching collections, classes or instances of the same. The proposed implementation should enable agent to also create,delete or update the graph according to the server changes via HTTP requests on endpoints. Did I miss anything?

Guttz:
The Hydra Spec  documents on its Endpoints which operations are available on them, for example, a /api/user endpoint may have the documentation for the POST operation and the GET operation. Right now, the client only is able to parse and execute the GET operations described on a hydrus server. 

The goal is to make the client be able to support all Hydra Supported operations GET, POST, PUT, DELETE and PATCH... That means, implement modifications so the Client is able to parse the all the operations from the Hydra Doc and also extend the CLI interface for the user to use it.

Mec-iS:
> enable the agent to also create,delete or update the graph according to the server changes via HTTP requests on endpoints

Yes, this is what we meant when we talk about CRUD. Every CRUD operation is usually mapped to an HTTP method. In the case of hydrus we decided this mapping:
* READ = GET
* UPDATE = POST
* CREATE = PUT
* DELETE = DELETE

Mec-iS:
@Guttz 
We need a more structured approach. Probably starting from a design document would be better.
1. We need first a list of requirements, we should have something like that already somewhere. We should map this requirements to [Patterns](https://python-patterns.guide/)
2. We need the underlying implementation of methods to interact with the server, conventions are established in hydrus for naming and factorizing functions. Base implementation should be [`requests.sessions.Session`](https://github.com/kennethreitz/requests/blob/master/requests/sessions.py#L340). Basically we want to do something like:
```
from requests.session import Session

class HydraAgent(Session):
    ...
```
`Session` has already all the methods for CRUDs
3. Once we have a basic client that can work in querying hydrus, we can create a parser for NLP that use regular expressions to read NLP commands

[...]

PR #105, opened 2019-05-31
Adding CREATE support

Guttz:
First step into the implementation of https://github.com/HTTP-APIs/hydra-python-agent/issues/101, a full featured client

**Also aligning develop branch which was behind**

### Checklist

### Description

Now can push to hydrus using the command: `create in <collection_endpoint> <json-ld object>`

### Change logs
Added insert in CollectionEndpoints class
Added create interpreter for cli
Added insert func in querying mech

#### Testing 
For testing this with the drone API just run:
`create in DroneCollection {"@type":"Drone","DroneState" : "Simplified state", "name" : "Smart Drone", "model" : "Hydra Drone", "MaxSpeed" : "999", "Sensor" : "Wind"}
`

### DIFF
--- a/hydra_agent/collections_endpoint.py
+++ b/hydra_agent/collections_endpoint.py
@@ -3,10 +3,11 @@ import json
 import re
 import logging
 from urllib.error import URLError, HTTPError
-from hydra_agent.classes_objects import ClassEndpoints,RequestError
+from hydra_agent.classes_objects import ClassEndpoints, RequestError
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+
 
 class CollectionEndpoints:
     """Contains all the collections endpoints and objects"""
@@ -23,13 +24,13 @@ class CollectionEndpoints:
         try:
             response = urllib.request.urlopen(new_url)
         except HTTPError as e:
-            logger.info('Error code: ', e.code)
+            logger.info('Request Error Code: ', e.code)
             return RequestError("error")
         except URLError as e:
-            logger.info('Reason: ', e.reason)
+            logger.info('Request Error Reason: ', e.reason)
             return RequestError("error")
         except ValueError as e:
-            logger.info("value error:",e)
+            logger.info("Request Value Error:", e)
             return RequestError("error")
         else:
             return json.loads(response.read().decode('utf-8'))
@@ -59,7 +60,6 @@ class CollectionEndpoints:
         :param url: Base url given by user.
         :param redis_connection: connection of Redis memory.
         """
-        print("accesing the collection object like events or drones")
         if endpoint_list:
             clas = ClassEndpoints(self.redis_graph, self.class_endpoints)
             for endpoint in endpoint_list:
@@ -76,7 +76,7 @@ class CollectionEndpoints:
                                               match_obj.group(2))
                 entrypoint_member = endpoint["@type"].lower(
                 ) + match_obj.group(3)
-#                print(base_url, entrypoint_member,endpoint["@type"])
+                # print(base_url, entrypoint_member,endpoint["@type"])
                 member_alias = entrypoint_member
                 # key for the object node is memeber_alias
                 member_id = match_obj.group(3)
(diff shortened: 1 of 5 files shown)


hydra_agent/collections_endpoint.py
```diff
@@ -23,17 +24,43 @@ def fetch_data(self, new_url):
...
             return RequestError("error")
         except ValueError as e:
-            logger.info("value error:",e)
+            logger.info("Request Value Error:", e)
             return RequestError("error")
         else:
             return json.loads(response.read().decode('utf-8'))
 
+    def put_data(self, collection_url, new_object):
+        """Fetching data from the server
```
xadahiya: Update the Docstrings

hydra_agent/collections_endpoint.py
```diff
@@ -201,8 +224,38 @@ def load_from_server(
                 redis_connection.delete(key)
         # save the new data.
         self.redis_graph.commit()
-#        for node in self.redis_graph.nodes.values():
-#            print("\n",node.alias)
+        # for node in self.redis_graph.nodes.values():
+        # print("\n",node.alias)
+
+    def insert_server(
```
xadahiya: insert_server is not very descriptive please use better function names. Also, update the doctrings

hydra_agent/querying_mechanism.py
```diff
@@ -526,110 +539,132 @@ def user_query(self, query):
...
+        elif "create" in query:
+            query = query.replace("create", "").strip()
+            if "in" in query:
+                query = query.replace("in", "").strip()
+                collection = query.split(" ", 1)[0]
+                new_object = query.replace(collection, "").strip()
+                self.members = CollectionmembersQuery(self.api_doc,
+                                                        self.url,
+                                                        self.graph)
+                response = self.members.insert_server('DroneCollection', '{"@type":"Drone", "DroneState" : "dronestate", "name" : "Drone3", "model" : "model", "MaxSpeed" : "speed", "Sensor" : "sensor"}')
```
xadahiya: This is hardcoded. Perhaps you forgot to update this

hydra_agent/querying_mechanism.py
```diff
@@ -192,6 +192,19 @@ def get_members(self, query):
             print(self.connection.smembers("fs:endpoints"))
             return self.data_from_server(endpoint)
 
+    def insert_server(self, endpoint, new_object):
```
xadahiya: What's the point of this function? You can directly call `insert_server` where you're calling this function.

hydra_agent/querying_mechanism.py
```diff
@@ -526,110 +539,132 @@ def user_query(self, query):
...
+                logger.info("Incorrect query: Use 'help' to know about\
+                            querying format")
                 return None
+        elif "create" in query:
+            query = query.replace("create", "").strip()
+            if "in" in query:
+                query = query.replace("in", "").strip()
+                collection = query.split(" ", 1)[0]
+                new_object = query.replace(collection, "").strip()
+                self.members = CollectionmembersQuery(self.api_doc,
```
xadahiya: Not sure you need to use `self.members` here, each query is run one time so using a simple variable `members` will do.

Mec-iS:
As discussed in the call, this is not the right approach. We don't need to fix the actual scripts but come out with a design that can embed the current code we have. I close this PR and open a proper issue for the job.
Let's start over from this ---> https://github.com/HTTP-APIs/hydra-python-agent/issues/101#issuecomment-497685202

hydra_agent/collections_endpoint.py
Guttz: I fixed all the other comments on my local repo. Thanks for the notes @xadahiya 

Regarding the function names, I tried to follow the previous standard. But guess I'll refactor them with the new design we discussed on the call

motormouthdoda:
I think you should start with improving the current GET mechanism rather than designing a new mechanism for PUSH/PUT. I think it'll help you avoid any contingencies that can occur later while designing other requests. Moreover improving the GET request will build up a base for other features as well (like advanced querying, better Redis integration, better structure for storing all the Hydra resources)

Guttz:
Yes, we're going with a design document and what will probably lead to refactoring the existing GET mechanism into this new architecture. Lorenzo shared more details at https://github.com/HTTP-APIs/hydra-python-agent/issues/101#issuecomment-497685202.

Next week I should be sharing a doc/design where you're welcome to give your inputs @motormouthdoda

PR #109, opened 2019-06-11
Agent: beginning implementation

Mec-iS:


hydra_agent/requests.py
```diff
@@ -0,0 +1,100 @@
...
+from redis_proxy import RedisProxy
+from graphutils import GraphUtils
+from redisgraph import Graph, Node
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class Requests:
+
```
Mec-iS: we don't need to reimplement this from scratch, use the library [`requests` 3](https://3.python-requests.org/).
Add `requests` to the requirements, then you just need:
```
from requests import Session

class Agent(Session):
    ....
```
This way you will have all the verbs and session management already implemented as [defined here](http://3.python-requests.org/api/#request-sessions).

First rule: we don't reinvent the wheel, always look for an existing popular and maintained library.

hydra_agent/requests.py
```diff
@@ -0,0 +1,100 @@
...
+    def __init__(self, entrypoint_url, redis_connection):
+        self.entrypoint_url = entrypoint_url
+        self.redis_connection = redis_connection
+        self.connection = redis_connection.get_connection()
+        self.vocabulary = 'vocab'
+        self.graph_utils = GraphUtils(redis_connection)
+        self.redis_graph = Graph("apidoc", redis_connection)
+
+    def get(self, url):
```
Mec-iS: As above, already implemented in `requests.Session.get`. This and all the other verbs just need to be overridden.

hydra_agent/requests.py
Guttz: Yes, agree. I was planning to do something like this:

requests.py:

```
from requests import Session

class Requests(Session):
    def get():

```


agent.py:
```
from hydra_agent.requests import Requests

class HydraAgent(Requests):
    ....

```

hydra_agent.requests will inherit requests3 session and already encapsulate basic Redis Interactions. 

agent will inherit hydra_agent.requests and have everything that session supports with basic Redis interactions already embedded in GET, POST and so on.. Do you agree?

Mec-iS: for bringing Redis in the picture you can use multiple inheritance or "mixin":
```
class Agent(Session, RedisProxy):
     ...
```
be careful about precedence in Python classes.

Mec-iS: https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-are-they-useful

Mec-iS: in particular our use-case is listed in this comment https://stackoverflow.com/questions/533631/what-is-a-mixin-and-why-are-they-useful#comment27661661_547714

Guttz: Okay, I'm studying the options and will experiment

Guttz:
Take a look @Mec-iS. I found here the opportunity to couple the Sync mechanism and I personally like the structure much more. Agent inherits Session now and calls SyncMechanism who is responsible to update Redis.

Also added DELETE in a similar way. 

Agent now declares, for example:


```
    def delete(self, url):
        response = super().delete(url)
        self.sync_processing.add_operation("DELETE", url)

        return response
```
And sync_processing has all the processing to update redis accordingly, the same method will be used for when a new Row is received from the Server's Modification Table.

I read about multiple inheritance and started to do it but by doing like above didn't feel the need to. Tell me what you think.

hydra_agent/agent.py
```diff
@@ -0,0 +1,32 @@
...
+from hydra_agent.redis_proxy import RedisProxy
+from hydra_agent.sync_mechanism import SynchronizationProcessing
+from requests import Session
+
+
+class HydraAgent(Session):
+    def __init__(self, entrypoint_url):
+        self.entrypoint_url = entrypoint_url
+        self.redis_proxy = RedisProxy()
+        self.sync_processing = SynchronizationProcessing(entrypoint_url,
```
Mec-iS: Yea it is fine. Instead of using inheritance we use composition/wrapping of an instance from another class

Mec-iS: I would prefer a function for every verb: `self._sync_get(...)` so it is more controllable and we avoid having `if...else` blocks for verbs.

Guttz: There is a function for every verb at [sync_mechanism.py ](https://github.com/HTTP-APIs/hydra-python-agent/pull/109/files#diff-22bb81c6fd017d49156ccfa7a9a90046)(sync_get(...), sync_delete(...))...  add_operation basically calls the right function.

I added this layer between(add_operation) because that's the way the information is going to be received from the Modification Table probably: | METHOD | RESOURCE_url | ...

Guttz:
This is the first structure and there are quite some things left, I mapped what I see as next steps:

#### Pre-processing for requests:

PUT: Check the apidoc to double-check and adjust object if necessary. Also, this has to support setting the id for the resource [ by IRI as in the spec](https://github.com/HydraCG/cookbook/blob/master/drafts/5.1.creating-event-with-put.md#usage). I started talking to @vddesai1871 about this.

POST: To support partial updates, we should fetch the full object and update the modified fields, sending back the full updated object.

GET: Basic check of seeing if it already exists on Redis

#### Auth
Should add support for auth by simply overriding the Session so it not only sets the auth for the Agent but also for the SyncProcessing Session(_Is this the right way? Look for suggestions)_

#### Layout
Also, will organize the project layout as we talked after the above basic working version.

Guttz:
[From conf call] Graph operations should be isolated from sync_mechanism. Renaming it to graphutils_operation.py.

Later sync_mechanism will call it, instantiate sockets and so on.

Mec-iS:
yea we are talking about two different layers: the sync_mechanism is a connectivity layer between server and client, the graphutils_ops perform in the data layer instead. Both the agent class and the sync_mechznism will use graphutils_ops

Mec-iS:
We have enough code to start writing unittest, @Guttz please get familiar with `unittest.mock` in the standard library or `requests-mock` (third party library) to mock requests-responses.

hydra_agent/graphutils.py
```diff
@@ -0,0 +1,95 @@
...
+            query += " WHERE(p.{})".format(where)
+        query += " RETURN p.{}".format(ret)
+
+        return self.redis_connection.execute_command("GRAPH.QUERY",
+                                                     self.graph_name,
+                                                     query)
+
+    def update(self, match, set, where=None):
+        """
+        Run query to read nodes in Redis and return the result
```
Mec-iS: wrong docstring

hydra_agent/graphutils.py
```diff
@@ -0,0 +1,95 @@
...
+        query = "MATCH(p:{})".format(match)
+        if where is not None:
+            query += " WHERE(p.{})".format(where)
+        query += " SET p.{}".format(set)
+
+        return self.redis_connection.execute_command("GRAPH.QUERY",
+                                                     self.graph_name,
+                                                     query)
+
+    def addNode(self, label1, alias1, properties1):
```
Mec-iS: remove the "1"s from signature

hydra_agent/graphutils.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self, redis_proxy, graph_name='apidoc'):
+        self.redis_proxy = redis_proxy
+        self.redis_connection = redis_proxy.get_connection()
+        self.graph_name = graph_name
+        self.redis_graph = Graph("apidoc", redis_proxy)
+
+    def read(self, match, ret, where=None):
+        """
+        Run query to read nodes in Redis and return the result
+        :param match: Relationship between queried entities.
```
Mec-iS: params shall have types, i.e.:
* `:param match (<type>): <description>`

hydra_agent/querying_mechanism.py
```diff
@@ -672,22 +672,20 @@ def main():
     Take URL as an input and make graph using initilize function.
     :return: call query function for more query.
     """
-    url = input("url>>>").strip()
+    url = input("url>>>").strip().rstrip('/')
     if url == "exit":
         print("exit...")
```
Mec-iS: No `print` statements are allowed in code. Code shall use Python standard library's `logging` package:
```
import logging

logger = logging.getLogger(__module__)
logger.info('This is a warning')
```
Logger can be initialised in every module or in the `__init__` of the package

hydra_agent/querying_mechanism.py
```diff
@@ -672,22 +672,20 @@ def main():
     Take URL as an input and make graph using initilize function.
     :return: call query function for more query.
     """
-    url = input("url>>>").strip()
+    url = input("url>>>").strip().rstrip('/')
```
Mec-iS: this is still using the CLI input, the function has to be made generic and eventually `click` has to be added for the CLI usage (we can add CLI functionalities later, now focus on generic functions)

hydra_agent/sync_mechanism.py
```diff
@@ -0,0 +1,153 @@
+import urllib.request
+import json
+import logging
+from urllib.error import URLError, HTTPError
+from redis_proxy import RedisProxy
+from graphutils import GraphUtils
+from redisgraph import Graph, Node
+from requests import Session
+
+logging.basicConfig(level=logging.INFO)
```
Mec-iS: `basicConfig` has to be declared once at the top of the package, then every module/package can declare its own `getLogger`

hydra_agent/graphutils.py
Guttz: Thanks for all the reviews, will be picking them up.

hydra_agent/graphutils.py
Guttz: I took a look at how it's done in hydrus and it's currently something like this:

image

Should I proceed with them like this or put them in the docstring?

hydra_agent/querying_mechanism.py
Guttz: Okay, will look out for that. Already replaced all the ones I had in my code, these are prints from the previous implementation on the nlp part. Will make sure to not use in all the new modules and I think I'll leave the refactoring of the nlp part when we get there.

hydra_agent/sync_mechanism.py
Guttz: I removed basicConfig and I'm using as we're on hydrus on top the module:

`logger = logging.getLogger(__file__)`

Is it okay like that?

Mec-iS: For now is ok, we will fix all the client configuration when addressing the proper issue

hydra_agent/graphutils.py
Mec-iS: Right, better use type annotations

hydra_agent/graphutils_operations.py
```diff
@@ -0,0 +1,147 @@
...
+            collection_members.append({'@id': resource['@id'],
+                                       '@type': resource['@type']})
+
+            # Updating the collection properties with the nem member
+            self.graph_utils.update(
+                match="collection",
+                where="id='{}'".format(redis_collection_id),
+                set="members = \"{}\"".format(str(collection_members)))
+            # Creating node for new collection member
+            node = self.graph_utils.add_node("objects" + resource['@type'],
```
Guttz: Spent some time into this but for some reason couldn't get it to set properly the relation and properties in the graph yet. It just doesn't add the properties correctly. Will comment on the solution here soon.

README.md
```diff
@@ -139,6 +139,48 @@ Query test can be done like this:
 
 For more detail take a look at [wiki file](https://github.com/HTTP-APIs/http-apis.github.io/blob/master/hydra-agent-redis-graph.md)
 
+#### Querying Redis
```
Guttz: I started this session with some important information, mainly the naming of labels and alias. I.e: If you want to find Drones by their label you have to:

`GRAPH.QUERY apidoc "MATCH (p:objectsDrone) RETURN p"` **p:Drone wouldn't work.**

I'll adjust, complement and write it better while I progress. For now, it's mainly for the sake of documenting as I go and I think it shouldn't even be on the main readme.md, perhaps a readme.md inside the Redis folder because it's mainly for our development use.

Mec-iS: we store the ApiDoc in the graph?

hydra_agent/graphutils_operations.py
Guttz: https://github.com/HTTP-APIs/hydra-python-agent/issues/110

README.md
Guttz: No, apidoc is the name for the Graph that Sandeep used in his code, basically the Redis key that has "graphdata" as its type. I guess it's because he started implementing having the apidoc in mind and later they changed... I guess this should also be changed to avoid confusion?

hydra_agent/graphutils_operations.py
Guttz: To properly create valid edges(relations) between existing nodes in RedisGraph(and not create nodes again) one should:
```
MATCH (f:%s{%s:'%s'}), (t:%s{%s:'%s'}) CREATE (f)-[:in]->(t)
```
With DroneCollection example:
```
GRAPH.QUERY apidoc "MATCH (s:collection {type:'DroneCollection'} ), (d:objectsDrone {id:'/serverapi/DroneCollection/ea7e438e-a93d-436d-a7e9-994c13d49dc0'} ) CREATE (s)-[:has_Drone]->(d)"
```
This is not available yet on the oficial doc and [one issue is open about it in RedisGraph repo](https://github.com/RedisGraph/redisgraph-py/issues/16). Will add to our internal RedisGraph doc that I'm populating.

README.md
Mec-iS: I suggest to change the name to `apigraph` or we can get confused

hydra_agent/graphutils_operations.py
Mec-iS: the issue states that `MATCH (f:%s{%s:'%s'}), (t:%s{%s:'%s'}) CREATE (f)-[:in]->(t)` is working though with some memory cleanup problem.

Guttz: Yeah.. I've been finding RedisGraph really awesome, think it fits really well with the Agent and will enable powerful things in the future. However, like Hydra, it's not really a finished product and has some limitations and enhancements on the way. Linking existing two nodes is somewhat a basic feature but still has these bits of loose ends/not so straight forward as desired.

Also, checking the graph for modifications during implementation takes a lot of query writing and time. They're still working on a visualization tool as mentioned in this issue https://github.com/RedisGraph/RedisGraph/issues/35 which, depending on the way it gets implemented, might be a stepping stone for a more visual Demo for the Agent. And ofc saving time when developing changes to the Agent and checking what's actually happening.

hydra_agent/graphutils_operations.py
```diff
@@ -0,0 +1,144 @@
...
+                                             resource_endpoint + "\'",
+                                             relation_type="has_" +
+                                             resource['@type'],
+                                             label_dest="objects" +
+                                             resource['@type'],
+                                             where_dest="id : \'" +
+                                             resource['@id'] + "\'")
+            return
+        except ValueError as e:
+            # Second Case - When processing a GET for a Colletion
```
xadahiya: Typo here, should be collection

hydra_agent/graphutils_operations.py
```diff
@@ -0,0 +1,144 @@
...
+    def get_processing(self, url: str, resource: dict) -> None:
+        """Synchronize Redis upon new GET operations
+        :param url: Resource URL to be updated in Redis.
+        :param resource: Resource object fetched from server.
+        :return: None.
+        """
+        url = url.rstrip('/').replace(self.entrypoint_url, "EntryPoint")
+        # Updating Redis
+        # First case - When processing a GET for a resource
+        try:
```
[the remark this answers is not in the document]
xadahiya: The you can encapsulate the whole thing in a try except block ( if you want) for logging purposes.

hydra_agent/graphutils.py
```diff
@@ -0,0 +1,120 @@
+from redisgraph import Node, Edge, Graph
+from typing import Union, Optional
+from redis.exceptions import ResponseError
+
+
+class GraphUtils:
+
+    def __init__(self, redis_proxy, graph_name='apidoc'):
```
xadahiya: Do we really want the graph name to be `apidoc`? This might create confusion

hydra_agent/graphutils.py
```diff
@@ -0,0 +1,120 @@
...
+        """Add edge between nodes in redis graph
+        :param source_node: source node of the edge.
+        :param predicate: relationship between the source and destination node
+        :param dest_node: destination node of the edge.
+        """
+        edge = Edge(source_node, predicate, dest_node)
+        self.redis_graph.add_edge(edge)
+
+    def commit(self) -> None:
+        """Commit the changes made to the Graph to Redi"""
```
xadahiya: Typo ( should be redis)

hydra_agent/graphutils_operations.py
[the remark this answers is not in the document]
Guttz: Makes sense, thanks for the suggestions. I will make a modification.

hydra_agent/graphutils.py
Guttz: > I suggest to change the name to apigraph or we can get confused

Yeah, Lorenzo had the same query, I was just following the name Sandeep used but I agree on changing it. I'll make the modification everywhere where it's necessary soon.

hydra_agent/agent.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if response.status_code == 200:
+            self.graph_operations.delete_processing(url)
+
+        return response
+
+if __name__ == "__main__":
+    Agent = HydraAgent("http://localhost:8080/serverapi")
+
+    new_object = {"@type": "Drone", "DroneState": "Simplified state",
```
xadahiya: All this should be kept in a separate module (you can create an examples folder if you want)

xadahiya: @Guttz fix this then we can merge

Guttz: I'll write some examples on the readme.md and remove these from main, sounds good?

Guttz: Chris suggested I finish some basic testing before merging. I think I'll upload one test later today so you guys can tell me if I'm going in the right direction and I move with the rest

hydra_agent/agent.py
```diff
@@ -0,0 +1,92 @@
...
+        self.initialize_graph()
+
+    def initialize_graph(self):
+        self.graph = InitialGraph()
+        self.redis_connection.delete("apigraph")
+        self.graph.main(self.entrypoint_url, self.api_doc, True)
+        self.redis_connection.sadd("fs:url", self.entrypoint_url)
+
+    def get(self, url):
+        response = self.graph_operations.get_resource(url)
```
Guttz: This is the first implementation to start using Redis, this gets already cached GET requests and return them. This avoids also duplications of getting the same resources into Redis more than once. The same function in the near future should receive somehow a property and also query by it.

Mec-iS:
I agree that we should have some minimal unit tests otherwise we cannot merge

hydra_agent/agent.py
```diff
@@ -0,0 +1,68 @@
...
+            return response
+
+        response = super().get(url)
+
+        if response.status_code == 200:
+            self.graph_operations.get_processing(url, response.json())
+
+        return response.json()
+
+    def put(self, url, new_object):
```
Guttz: The agent is missing docstrings and I'll complement some signature types also

Mec-iS: yes thanks. after that I am ok to merge

xadahiya:
@Guttz Ready to merge?

Issue #113, opened 2019-06-28
Recursive node creation


Guttz:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

As detailed in https://github.com/HTTP-APIs/hydra-python-agent/issues/110, currently the agent inserts nodes and create the edges for all resources fetched, but it doesn't recursively do the same for all the URIs that may be embedded in the resource's properties.

### Expected Behaviour:

The agent should create nodes and links to all URIs embedded in the resource's properties

### Do you want to work on this issue?

Yes, I think this is one of the next steps

Issue #112, opened 2019-06-28
Redis Graph py Result-set format update


Guttz:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Current Behaviour:

Our repo is currently using redisgraph-py:1.7 and there's a breaking change to 2.0 because they've changed the Result-set into a more easily processable output.. Which might be interesting.

### Mailing update and ref:
https://groups.google.com/forum/#!topic/redisgraph/UHC_f3a7Ldo
https://oss.redislabs.com/redisgraph/result_structure/

### Modifications to be made:
redisgraph-py - 1.7 -> 2.0 https://github.com/RedisGraph/redisgraph-py/releases 
redisgraph docker image -  1.2.2 -> "edge"  https://hub.docker.com/r/redislabs/redisgraph/tags

xadahiya:
We should deal with this on a priority basis then.

Mec-iS:
yea we cannot get stuck to the old version at this point

Guttz:
Yup, I agree with that. Will try to make the upgrade for the repo.

Guttz:
Almost finished updating, the new response format is better and already does some code I had to program.

However, there are still some loose things, opened an issue over there -> https://github.com/RedisGraph/redisgraph-py/issues/42. Will clear some things out before moving on with this.

PR #118, opened 2019-07-03
Recursive node creation

Guttz:
Fixes #
https://github.com/HTTP-APIs/hydra-python-agent/issues/113

### Checklist

### Description

Now the agent create nodes and links to all URIs embedded in a resource property when finding one

### Discussion
One thing that popped in my mind and I couldn't find an answer is about the name of the edge. I.e:

StateCollection -> has_State -> State1
Drone123123 -> has_State -> State1

Should we define a different naming for edges between collections-members and members-members?

### DIFF
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -49,7 +49,17 @@ class Agent(Session):
         response = super().get(url)
 
         if response.status_code == 200:
-            self.graph_operations.get_processing(url, response.json())
+            # Graph operations returns embedded resources if finding any
+            embedded_resources = \
+                self.graph_operations.get_processing(url, response.json())
+            # Embedded resources are fetched and then properly linked
+            for embedded_resource in embedded_resources:
+                self.get(embedded_resource['embedded_url'])
+                self.graph_operations.embedded_resource(
+                    embedded_resource['parent_id'],
+                    embedded_resource['parent_type'],
+                    embedded_resource['embedded_url']
+                )
             return response.json()
         else:
             return response.text
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -11,6 +11,9 @@ logger = logging.getLogger(__file__)
 
 
 class Agent(Session):
+    """Provides a straightforward GET, PUT, POST, DELETE -
+    CRUD interface - to query hydrus
+    """
     def __init__(self, entrypoint_url: str) -> None:
         """Initialize the Agent
         :param entrypoint_url: Entrypoint URL for the hydrus server
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -55,13 +55,7 @@ class Agent(Session):
             # Graph_operations returns the embedded resources if finding any
             embedded_resources = \
                 self.graph_operations.get_processing(url, response.json())
-            # Embedded resources are fetched and then properly linked
-            for embedded_resource in embedded_resources:
-                self.get(embedded_resource['embedded_url'])
-                self.graph_operations.link_resources(
-                    embedded_resource['parent_id'],
-                    embedded_resource['parent_type'],
-                    embedded_resource['embedded_url'])
+            self.process_embedded(embedded_resources)
             return response.json()
         else:
             return response.text
@@ -79,13 +73,7 @@ class Agent(Session):
             # Graph_operations returns the embedded resources if finding any
             embedded_resources = \
                 self.graph_operations.put_processing(url, new_object)
-            # Embedded resources are fetched and then properly linked
-            for embedded_resource in embedded_resources:
-                self.get(embedded_resource['embedded_url'])
-                self.graph_operations.link_resources(
-                    embedded_resource['parent_id'],
-                    embedded_resource['parent_type'],
-                    embedded_resource['embedded_url'])
+            self.process_embedded(embedded_resources)
             return response.json(), url
         else:
             return response.text, ""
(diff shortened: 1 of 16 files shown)


hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -5,32 +5,36 @@
 from hydra_agent.redis_core.redis_proxy import RedisProxy
 from hydra_agent.redis_core.graphutils import GraphUtils
 from redisgraph import Graph, Node
+from requests import Session
 
 logger = logging.getLogger(__file__)
 
 
 class GraphOperations():
```
Mec-iS: missing docstring

hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -5,32 +5,36 @@
...
+    def __init__(self, entrypoint_url: str, api_doc: dict,
+                 redis_proxy: RedisProxy):
         self.entrypoint_url = entrypoint_url
+        self.api_doc = api_doc
         self.redis_proxy = redis_proxy
         self.redis_connection = redis_proxy.get_connection()
         self.vocabulary = 'vocab'
         self.graph_utils = GraphUtils(redis_proxy)
         self.redis_graph = Graph("apigraph", self.redis_connection)
+        self.session = Session()
```
Mec-iS: `Session` shouldn't be called here. For any call that needs a session the code should call a function in the agent class. So probably you wan to create a method in the other class, and make this class to call it. In the current way, segregation of concerns is violated.

hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -192,5 +211,36 @@ def get_resource(self, url: str) -> dict:
...
+                          discovered_url: str) -> str:
+        """Checks for existance of discovered resource and creates links
+        for embedded resources inside other resources properties
+        :parent_id: Resource ID for the parent node that had this reference
+        :parent_type: Resource Type for the parent node that had this reference
+        :discovered_url: URL Reference for resource found inside a property
+        """
+        resource = self.get_resource(discovered_url)
+        if resource is None:
+            response = self.session.get(discovered_url)
```
Mec-iS: As written above, this line should call a method from another class, not make this class to instantiate its own Session.

hydra_agent/redis_core/graphutils_operations.py
Guttz: You are right, this would get messed up in the future. But I cannot simply instantiate a new Agent and call it, because it will be a new session and not inherit the Agent session's attributes.

What's the right way to call the Agent from here persisting the session since this module is instantiated by the Agent?

Mec-iS: Session are really volatile. you can have a classmethod in Agent like `Agent.get_session()`, that just instantiate and return a session, without assigning to `self.session`. I meant, use an helper, do not assign a session to the operations instance, otherwise every instance will need to carry around a session without needing it. Factorise the functionality, so to avoid to add the functionality to all the instances.

Guttz: Okay, I also researched some design patterns and got some ideas. I'll just finish up a PR about updating Redis graph for #112 and then I'll propose a solution for this.

hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -40,13 +52,13 @@ def get_processing(self, url: str, resource: dict) -> None:
             collection_members = self.graph_utils.read(
                 match=":collection",
                 where="id='{}'".format(redis_collection_id),
-                ret=".members")
+                ret="")
 
             # Checking if it's the first member to be loaded
-            if collection_members is None:
+            if not 'members' in collection_members:
```
vddesai1871: I think replacing it with `if 'members' not in collection_mambers` would be more readable.

hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -141,13 +173,13 @@ def delete_processing(self, url: str) -> None:
         collection_members = self.graph_utils.read(
             match=":collection",
             where="id='{}'".format(redis_collection_id),
-            ret=".members")
+            ret="")
 
         # Checking if it's the first member to be loaded
-        if collection_members is None:
-            return
+        if not 'members' in collection_members:
```
vddesai1871: same here

hydra_agent/redis_core/graphutils_operations.py
```diff
@@ -191,5 +223,33 @@ def get_resource(self, url: str) -> dict:
 
         return resource
 
+    def link_resources(self, parent_id: str, parent_type: str,
+                       node_url: str) -> str:
+        """Checks for existance of discovered resource and creates links
+        for embedded resources inside other resources properties
+        :parent_id: Resource ID for the parent node that had this reference
+        :parent_type: Resource Type for the parent node that had this reference
+        :node_url: URL Reference for resource found inside a property
```
vddesai1871: Documentation of return value is missing.

hydra_agent/redis_core/graphutils_operations.py
Guttz: I did this for all occurrences in the PR I'm opening soon, the other way is actually not pep8 conformant

hydra_agent/agent.py
```diff
@@ -47,9 +52,19 @@ def get(self, url: str) -> Union[dict, list]:
         response = super().get(url)
 
         if response.status_code == 200:
-            self.graph_operations.get_processing(url, response.json())
-
-        return response.json()
+            # Graph operations returns embedded resources if finding any
```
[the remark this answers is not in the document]
Guttz: Yep, implemented the helper and the code looks much cleaner.

I fixed all the other typos, but in this case "Graph_operations(**it, the module**) returns", think it's right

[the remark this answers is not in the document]
vddesai1871: Yeah obviously.

PR #121, opened 2019-07-11
Server searching by type and properties

Guttz:
### Checklist

### Description

Agent will search both Redis and Server for resources by type and properties. Ex.: 
print(agent.get(resource_type="Drone", filters={'name': "Drone 2"}))

### Functionality 

- Resources queried by URL will be fetched directly from Redis. If they are not yet cached, they will be fetched from the server

- Searching by type and properties will fetch from Redis considering the optional variable "cached_redis" that sets the minimum limit of cached resources you would like. If not enough resources, the search will be forward to the server. By default, for integrity purposes, it's set for sys.maxsize.

- Collections will always be fetched from Server since new members(unavailable in Redis) can always be added

### DIFF
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -41,26 +41,40 @@ class Agent(Session):
         self.redis_connection.sadd("fs:url", self.entrypoint_url)
 
     def get(self, url: str = None, resource_type: str = None,
-            filters: dict = {}) -> Union[dict, list]:
+            filters: dict = {}, cached_limit: int = 10) -> Union[dict, list]:
         """READ Resource from Server/cached Redis
         :param url: Resource URL to be fetched
+        :param resource_type: Resource object type
+        :param filters: filters to apply when searching, resources properties
+        :param cached_limit : Minimum amount of resources to be fetched
         :return: Dict when one object or a list when multiple targerted objects
         """
-        response = self.graph_operations.get_resource(url, resource_type,
-                                                      filters)
-        if response:
-            return response
-        elif url is None:
-            return []
+        redis_response = self.graph_operations.get_resource(url, resource_type,
+                                                            filters)
+        if redis_response:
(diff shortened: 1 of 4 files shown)


hydra_agent/agent.py
```diff
@@ -41,26 +41,41 @@ def initialize_graph(self) -> None:
         self.redis_connection.sadd("fs:url", self.entrypoint_url)
 
     def get(self, url: str = None, resource_type: str = None,
-            filters: dict = {}) -> Union[dict, list]:
+            filters: dict = {},
+            cached_limit: int = sys.maxsize) -> Union[dict, list]:
```
xadahiya: Not sure if using `sys.maxsize` as the cached_limit is a good idea here.
From StackOverflow

> Python can handle arbitrarily large integers in computation. Any integer too big to fit in 64 bits (or whatever the underlying hardware limit is) is handled in software. For that reason, Python 3 doesn't have a sys.maxint constant.
The value sys.maxsize, on the other hand, reports the platform's pointer size, and that limits the size of Python's data structures such as strings and lists.

What are we trying to achieve with this?

Guttz: This was a way to use the Redis cached data for collections and search by type/classes, so the user wouldn't always query the server. Maxsize is a way to set that by default it will always fetch the server because Redis won't have "Maxsize" amount of nodes, which then allows the user to control by itself.

Now that we have https://github.com/HTTP-APIs/hydrus/pull/409 I could check for the pagination limit and return only the cached data if enough, but if the user is querying multiple pages this might lead to inconsistency problems because the cached data won't be in the same order.

Removing this functionality is also an option, then Redis use case will be mainly on getting specific collection members and its properties instead of going for the server.

Mec-iS: Let's go with your implementation @Guttz 
we can double-check later eventually.

Issue #98, opened 2019-03-21
Create a synchronisation mechanism for server-client udpates


Mec-iS:
this couples with https://github.com/HTTP-APIs/hydrus/issues/300

Agent should be able to receive server-updates and update the graph.

PR #123, opened 2019-07-18
Synchronization Mechanism for Server-Client Updates

Guttz:
Fixes https://github.com/HTTP-APIs/hydra-python-agent/issues/98 (Agent) https://github.com/HTTP-APIs/hydrus/issues/300 (hydrus)

### Checklist

### Description

Agent now connected to a socket and keep internal resources updated avoiding the use of invalid data.

**The server core behavior was architected and simulated at https://github.com/Guttz/simulated-hydrus-sync-socket. Take a quick look at the Readme on how it works.**

### Brief Summary
- Server has an outbox table, table has a unique column "job_id". Agent keeps track of modifications to make using the last updated "job_id"
- Agent connects to the Socket, copy the table and listen for updates
- When modifications are made to hydrus, it emits an event to the clients
- Clients query endpoint modification-table-diff, fetch new rows and update resources accordingly

Agent follows **almost exactly** what's described at:
https://github.com/Guttz/simulated-hydrus-sync-socket/blob/master/app.py

The only modification is that, instead of using Dates for resources merging, now hydrus table is the only source of truth and order must be followed as it's there for integrity and simplicity matters. That means that the agent always fetches blocks of missing rows instead of receiving each information alone and trying to merge by itself.

Another important note is that the Operations that are tracked are POST and DELETE since they are the only ones that can lead to integrity issues. GET requests from other clients do not change the Server state and PUT requests are new resources that the Client hasn't queried yet, so they aren't yet cached and should be ignored.

### DIFF
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -26,7 +26,7 @@ class Agent(Session):
         jsonld_api_doc = super().get(self.entrypoint_url + '/vocab').json()
         self.api_doc = doc_maker.create_doc(jsonld_api_doc)
         self.initialize_graph()
-        self.graph_operations = GraphOperations(entrypoint_url,
+        self.graph_operations = GraphOperations(self.entrypoint_url,
                                                 self.api_doc,
                                                 self.redis_proxy)
 
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -1,4 +1,5 @@
-import logging, sys
+import logging
+import sys
 import socketio
 from hydra_agent.redis_core.redis_proxy import RedisProxy
 from hydra_agent.redis_core.graphutils_operations import GraphOperations
@@ -42,7 +43,7 @@ class Agent(Session, socketio.ClientNamespace, socketio.Client):
         socketio.Client.connect(self, self.entrypoint_url_temp,
                                 namespaces=namespace)
         self.last_job_id = ""
-        self.modification_table = []
+        self.modifications_table = []
 
     def initialize_graph(self) -> None:
         """Initialize the Graph on Redis based on ApiDoc
@@ -159,10 +160,10 @@ class Agent(Session, socketio.ClientNamespace, socketio.Client):
         """Method executed when the Agent is successfuly connected to the Server
         """
         logger.info('Socket Connection Established - Synchronization ON')
-        self.modification_table = super().get(self.entrypoint_url_temp +
+        self.modifications_table = super().get(self.entrypoint_url_temp +
                                               '/modification-table-diff').json()
-        if self.modification_table:
-            self.last_job_id = self.modification_table[0]['job_id']
+        if self.modifications_table:
+            self.last_job_id = self.modifications_table[0]['job_id']
 
     def on_disconnect(self):
         """Method executed when the Agent is disconnected
@@ -180,10 +181,10 @@ class Agent(Session, socketio.ClientNamespace, socketio.Client):
         if not new_rows:
             logger.info('Server Restarting - Automatic Sync not possible')
             self.initialize_graph()
-            self.modification_table = super().get(self.entrypoint_url_temp +
+            self.modifications_table = super().get(self.entrypoint_url_temp +
                                                   '/modification-table-diff').json()
-            if self.modification_table:
-                self.last_job_id = self.modification_table[0]['job_id']
+            if self.modifications_table:
+                self.last_job_id = self.modifications_table[0]['job_id']
             return None
 
         # Checking if the Agent is too outdated and can't be synced
@@ -196,8 +197,8 @@ class Agent(Session, socketio.ClientNamespace, socketio.Client):
                     self.graph_operations.delete_processing(row['resource_url'])
 
         # Deleting old modifications table since they won't be used again
-        self.modification_table = new_rows
-        self.last_job_id = self.modification_table[0]['job_id']
+        self.modifications_table = new_rows
+        self.last_job_id = self.modifications_table[0]['job_id']
 
     def on_broadcast_event(self, data):
         """Method executed when the Agent receives a broadcast event
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -38,9 +38,7 @@ class Agent(Session, socketio.ClientNamespace, socketio.Client):
         socketio.ClientNamespace.__init__(self, namespace)
         socketio.Client.__init__(self, logger=True)
         socketio.Client.register_namespace(self, self)
-        # After hydrus implementation connect directly to the sv, self.entrypoint
-        self.entrypoint_url_temp = 'http://localhost:5000'
-        socketio.Client.connect(self, self.entrypoint_url_temp,
+        socketio.Client.connect(self, self.entrypoint_url,
                                 namespaces=namespace)
         self.last_job_id = ""
         self.modifications_table = []
--- a/hydra_agent/agent.py
+++ b/hydra_agent/agent.py
@@ -1,6 +1,5 @@
 import logging, sys
 import socketio
-import urllib.request
 from hydra_agent.redis_core.redis_proxy import RedisProxy
 from hydra_agent.redis_core.graphutils_operations import GraphOperations
 from hydra_agent.redis_core.graph_init import InitialGraph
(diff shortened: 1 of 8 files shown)


Guttz:
**Todo:**
- [ ] Implement Restart_Agent when it's too outdated, therefore not being able to Sync
- [ ] Add Docstrings and other refactorings

I've spent some time to make this work properly and make it concise, in the I've got it to work with a triple inheritance. If you have any suggestions on a better approach let me know.

xadahiya:
We still need to make hydrus compatible correct? In that case, please coordinate with vishal on this.

xadahiya:
Do we really need an inbox table? 
Why not keep things simple. Here's a simple disign that might work.
- First we implement socket functionality in hydrus
- Whenever any resource is added/deleted/updated then hydrus will broadcast this information to a socket broadcast channel
- When started, the agent will connect to this broadcast channel and will keep listening for messages
- If the information is relevant to the agent (based on the items agent has cached locally) then the agent will synchronize accordingly, if not, it can just ignore the message

This approach will work even if multiple client are connected to hydrus simultaneously and that too without much complexity.

Thoughts @vddesai1871 @Guttz @Mec-iS @chrizandr ^^

Mec-iS:
Yea probably for simplicity, in the beginning, is enough to have only an outbox deep enough in hydrus, so the client can double-check if all the changes are there. Still the client needs a way to check if all the changes have been applied in its local representation, so it needs to keep a log of a series of changes already done. This is technically the inbox, you can call it log, but still it is a needed part for matching the changed already applied by the client and the outbox of the server.

xadahiya:
> This is technically the inbox, you can call it "log" but it doesn't change the substance, but still it is a needed part for matching the changed already applied by the client and the outbox of the server.

Makes sense! But we'll need this at the client-side only  (at- least for now) correct? Since maintaining the history of all the connected clients and their syncs will be a lot of work on the server-side.
For now, I think we can ask the client to fetch all the resources again when reconnected.

vddesai1871:
> Here's a simple disign that might work.
Yes, I agree with this design. We might need to add some mechanism to support reconnects, so to tackle that as already mentioned by @Mec-iS  we can have a table at hydrus which stores details of specific kind of operation with timestamps, when client reconnects it can get log of all changes occurred after some specific time, hydrus returns those and client can update whichever nodes it cares about

vddesai1871:
>For now, I think we can ask the client to fetch all the resources again when reconnected.

It might work too, good enough to start

Mec-iS:
Yea that is what an inbox is for in a message-passing architecture

Guttz:
The implementation in the PR is almost what you described @xadahiya. The difference is what Lorenzo mentiod, I actually said it incorrectly. They don't hold the same inbox, hydrus has an outbox and all the client has is a table with its last synced updates(usually one row), and a variable call "last_job_id". This variable is used to fetch the new updates, the ones after that made after this job_id, and proceed with them.

I'll comment on the hydrus issue #300 soon the details of the hydrus. The simulation I shared has almost all everything, all clients connect and then server informs all clients when new updates were done, then the clients make a request to get the new updates/rows according to their last updated job_id.

> For now, I think we can ask the client to fetch all the resources again when reconnected.
hydrus only needs a limited history of its updates, for example, 100 or 1000 rows(we have to discuss a reasonable amount). He doesn't need to store the clients since he will send a notification for all the clients that there is new info. The client has a variable of the last synced job id, if losing connection or restarting the server, If the client has a last synced job ID that isn't on the table anymore, it will restart Redis and copy the current modifications_table so the Agent doesn't use outdated data.

xadahiya:
> I'll comment on the hydrus issue #300 soon the details of hydrus. The simulation I shared has almost everything, all clients connect and then server informs all clients when new updates were done, then the clients make a request to get the new updates/rows according to their last updated job_id.

Will take a look tonight :)

> hydrus only needs a limited history of its updates, for example, 100 or 1000 rows(we have to discuss a reasonable amount). He doesn't need to store the clients since he will send a notification for all the clients that there is new info. The client has a variable of the last synced job id, if losing connection or restarting the server, If the client has a last synced job ID that isn't on the table anymore, it will restart Redis and copy the current modifications_table so the Agent doesn't use outdated data.

Okay, Makes sense.

Guttz:
I just made a comment https://github.com/HTTP-APIs/hydrus/issues/300#issuecomment-513204331 on how hydrus would work in the scenario we are discussing.

hydra_agent/agent.py
```diff
@@ -140,5 +155,55 @@ def process_embedded(self, embedded_resources: list) -> None:
...
+    def on_disconnect(self):
+        """Method executed when the Agent is disconnected
+        """
+        pass
+
+    def on_update(self, data) -> None:
+        """Method executed when the Agent receives an event named 'update'
+        This is sent to all clients connected the server under the designed Namespace
+        """
+        new_rows = super().get(self.entrypoint_url_temp +
```
vddesai1871: I don't see the `data` parameter being used here. I think here we should primarily use the information provided in `data` to update the client instead of using a separate request to the server on `modification-table`. 
What I propose is
1. `data` parameter will contain an object with fields `last_job_id`, `new_job_id`, `operation_type` and `object_uri`. 
2. If `last_job_id` of client and `last_job_id` field of `data` matches then you can directly use `operation_type` and `object_uri` to make appropriate changes at the client and update client's `last_job_id` with `new_job_id`.
3. If they don't match then only client sends a request at `modification-table`.

Guttz: Well, that's much better actually. I agree we should proceed like this.

At first, I was avoiding using the data in the socket afraid that multiple clients would lead to tables having a different order(integrity problem), but putting "last_job_id" inside the socket data event will save a lot of requests and will fall into modification-table-diff only in eventual ocasions when necessary. Good solution! @vddesai1871  :+1:

hydra_agent/agent.py
```diff
@@ -10,26 +12,39 @@
...
+        self.graph_operations = GraphOperations(self.entrypoint_url,
                                                 self.api_doc,
                                                 self.redis_proxy)
 
+        # Declaring Socket Rules and instaciating Synchronization Socket
+        socketio.ClientNamespace.__init__(self, namespace)
+        socketio.Client.__init__(self, logger=True)
+        socketio.Client.register_namespace(self, self)
+        # After hydrus implementation connect directly to the sv, self.entrypoint
+        self.entrypoint_url_temp = 'http://localhost:5000'
```
xadahiya: Don't you think this shouldn't be hardcoded? Rest everything is looking good. 👍

hydra_agent/agent.py
```diff
@@ -140,5 +153,58 @@ def process_embedded(self, embedded_resources: list) -> None:
...
+        """Method executed when the Agent is disconnected
+        """
+        pass
+
+    def on_update(self, data) -> None:
+        """Method executed when the Agent receives an event named 'update'
+        This is sent to all clients connected the server under the designed Namespace
+        """
+        if data['last_job_id'] == self.last_job_id:
+            new_rows = [data]
```
xadahiya: Won't this always be one element? (See #406 in hydrus, L244 helpers.py)

hydra_agent/agent.py
Guttz: Yeah, that was weak memory, sorry. I fixed all connections to fetch the same link as the hydrus server, so it's working with Vishal's implementation now.

hydra_agent/agent.py
Guttz: Yeah it will, I did it like this to transform from a dict to a list of length 1 so the code below remains the same and can be used for both scenarios- fetching multiple updates or just this one from the socket.

Guttz:
The tests seem to be failing now due to some redis graph update, will take a look. The last commits in the past days are targeted towards formatting/coding standard and compatibility matters here and in the hydrus PR which Vishal and I are discussing.

xadahiya:
@Guttz See test logs here https://api.travis-ci.com/v3/job/221451656/log.txt

Guttz:
Since Server-side is merged now I'll double-check the functionality and fix the compatibility matter so it gets ready to merge

Guttz:
Redis graph compatibility fixed

xadahiya:
@guttz are you done with the checks?

Guttz:
Everything double-checked @xadahiya. Pushed a small modification to better handle a warning and now everything lgtm.

You can merge :+1:

Issue #124, opened 2019-08-04
Hydra Agent GUI Console


Guttz:
### I'm submitting a 
- [ ] bug report.
- [x] feature request.

### Description

The agent should have a GUI that enables the user to try and interact with a hydrus server through an interactive interface. Full demo reference mockup:  

![Group h1](https://user-images.githubusercontent.com/11739405/61073311-88d7d900-a3eb-11e9-9607-b97a4b3c0a65.png)

For now, as suggested and settled by @chrizandr in our weekly call, I will focus on the right side of the demo. Thus, the goal now is to give the Agent a 100% functional graphical console to interact with a deployed hydrus server. That means, being able to connect to a hydrus server and having a dynamic UI built based on the ApiDoc of the server that builds valid requests and let the user see the output. 

More info on https://github.com/HTTP-APIs/http-apis.github.io/issues/58

### Do you want to work on this issue?

Yes

Guttz:
One observation @xadahiya asked me to mention was about the framework decision. I've considered mainly two things:

1. The left side of the demo which has the graph which needs a stable and complete library for it not be a problem in the future while generating the Dynamic Graphs. I've checked and tested the best libraries(d3, visjs, sigma.js) and had a much easier time trying to implement what I wanted with React than with Flask. 

2. I have much more previous experiences in Javascript working with frontend and personally prefer it because of all the support around it. So I think this will be more effective considering our time frame and that I'll be responsible for the frontend.

In the end, I've liked vis.js the best for our goal, [it has an extensive library for networks
](https://visjs.github.io/vis-network/docs/network/)  and I've made it work with React and tested some basic graphs like the one below:

![image](https://user-images.githubusercontent.com/11739405/62428836-b78c5b00-b6dc-11e9-9dd1-e244943cd8c0.png)

xadahiya:
Sounds good 👍

xadahiya:
Moved to https://github.com/HTTP-APIs/hydra-python-agent-gui

PR #125, opened 2019-08-05
Hydra Agent Console Implementation 

Guttz:
Related to:  https://github.com/HTTP-APIs/hydra-python-agent/issues/124

Draft pull request that will continuously be updated along the next days


xadahiya:
Can you share some rough mockups first

Guttz:
@xadahiya Sure, today I'll share here where I am/next steps

Guttz:
Here's a picture of the current state:

(Click to expand)
image

This is the developed static version, with some small differences to the Mockup. I've added @xadahiya's suggestion, which is an input field that will contain the raw command. The first icon on the upper toolbar is to hide the HydraGraph(so I can focus only on the right as Chris suggested), the second leads to our home page. I've made one big design change from the mockup that was to use our yellow to match Hydra Eco's colors. The overall aspect should be this with some modifications to functionality as I progress with the actual coding. 

I uploaded  a small video so you can see everything: https://youtu.be/qiXgwdAltJE

Guttz:
#### Next steps
As for the next steps I plan to start to port each of the logic into modular dynamic pieces for it to be built based on the ApiDoc. For that, I'll create a Mock Flask backend(The Agent layer that will forward the commands) which will just return the Mocked processed object that the Frontend needs for now. Later when Vishal is able to join, I think implementing the Flask Endpoints with the actual processing replacing the Mocks would be a good task.

Guttz:
I've created the Flask backend with five endpoints: 
- serve(react app) 
- start-agent(receives hydrus url) 
- hydra-doc(sends hydra doc to frontend).

There are two endpoints I've created but the functionality is yet to be developed: 
- apidoc_graph(send a formatted representation of the api doc), I've already written the format there. 
- send_command(communication endpoint to forward messages to hydrus)
 
With that, I started implementing the dynamic components and the Collection Buttons are already generated based on the ApiDoc.

image

Guttz:
### Next steps

Will finish up the ` Collection buttons component `and proceed with the `operations component `and `properties component`, so they also are generated according to the ApiDoc. After that, I will code the interactions of building the request combining these three components.

xadahiya:
Porting the UI https://github.com/HTTP-APIs/hydra-python-agent-gui

Guttz:
Finished Dynamic buttons and ported it to a modular component. I was having some context problems but everything is ok now, the Console component can already receive the collection component clicked according to the apidoc:

ready

> Porting the UI https://github.com/HTTP-APIs/hydra-python-agent-gui
As we've settled on our last call Friday I'll port this GUI to another repo soon.

Guttz:
One observation @vddesai1871 asked me to raise about the api_doc endpoint on the Flask backend. We have one endpoint for a raw Hydra Doc(used to build the UI) and one for a processed Hydra Doc in Visjs's format(used to build the Graph). The second one could be done both front end and backend. I decided it would be better to do it backend for two reasons:

1. It's really likely that changes in Hydra/Core will be made in the future, and this processing might have to be updated. And as our codebase is mostly in Python, I think having it on the Flask Python Backend side than having it in jsx/React would make it much easier for future contributors to interact when it's necessary.
2. If we want to modify and extend the visualization to show some resources or some other information as a summary or connection to other servers in the future, this could be done only backend side and the Ui doesn't have to be modified and will already be compatible (

Guttz:
This PR was ported to a new repo as commented above.

A small PR was merged(https://github.com/HTTP-APIs/hydra-python-agent/pull/126) related to a small functionality here in the agent that is important for the GUI.

PR #164, opened 2021-03-13
Update requirements.txt

Mec-iS:
redisgraph and hydra_python_core are now on PyPi

It seems the version provided in PyPi is not the same provided from Github.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ redis
 graphviz
 requests
 python-socketio
--e git+https://github.com/RedisGraph/redisgraph-py#egg=redisgraph
--e git+https://github.com/http-apis/hydra-python-core@master#egg=hydra_python_core
+redisgraph
+hydra_python_core


requirements.txt
```diff
@@ -6,5 +6,5 @@ redis
 graphviz
 requests
 python-socketio
--e git+https://github.com/RedisGraph/redisgraph-py#egg=redisgraph
--e git+https://github.com/http-apis/hydra-python-core@master#egg=hydra_python_core
+redisgraph
```
priyanshunayan, line 9: We can't move ahead with the latest release of `redisgraph`. They have backwards incompatible change which breaks the agent. Currently we are depending upon the last working commit for us of redisgraph.(see requirements.txt of develop branch of agent). 

cc: @Mec-iS

Mec-iS: this should be written and noted somewhere. There is a comment or a note in README? We should have a PR open to fix it, it could be a good task for early contributors.

